PR #63, opened 2022-02-01
Feature/cognigy enhancements alex

akirilyuk:
add more cognigy features


lib/tasks/cognigy/index.js
```diff
@@ -22,7 +23,7 @@ const parseQuickReplies = (obj) => {
 
 const parseBotText = (evt) => {
   const {text, data} = evt;
-  if (text) return text;
+  if (text !== undefined) return String(text);
```
davehorton, line 26: what is the purpose of this?  Why are you not calling text.toString() instead of String as a function?

lib/tasks/cognigy/index.js
```diff
@@ -184,19 +252,57 @@ class Cognigy extends Task {
...
+    if (text) {
+      this._enqueueTask(async() => {
+        this.logger.info({text}, 'received text');
+        const sayTask = this._makeSayTask(text);
+        await sayTask.exec(cs, ep, this);
+      });
+    }
+
+
+    if (evt && evt.data && evt.data.type) {
```
davehorton, line 287: possible to remove this and just reference evt?.data?.type in the switch below?

lib/tasks/cognigy/index.js
akirilyuk: It could be that a text is a boolean, instead of a string or a number. In this case we need to explicitly parse it to string.

The explicit undefined check is implemented to also support check for a false value, since otherwise if text === false, we will not run in the if cause even if text is actually defined.

lib/tasks/cognigy/index.js
akirilyuk: Sure

PR #70, opened 2022-02-14
add say and gather task features

akirilyuk:
* add support for barge in in gather task
* add support of changing single say task TTS options on the fly

### DIFF
--- a/lib/tasks/say.js
+++ b/lib/tasks/say.js
@@ -21,15 +21,20 @@ class TaskSay extends Task {
     const {updateSpeechCredentialLastUsed} = require('../utils/db-utils')(this.logger, srf);
     const {writeAlerts, AlertType, stats} = srf.locals;
     const {synthAudio} = srf.locals.dbHelpers;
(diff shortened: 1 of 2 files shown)


lib/tasks/say.js
```diff
@@ -21,15 +21,26 @@ class TaskSay extends Task {
...
+      : cs.speechSynthesisLanguage ;
+    const voice =  this.synthesizer.voice && this.synthesizer.voice !== 'default'
+      ? this.synthesizer.voice
+      : cs.speechSynthesisVoice;
     const engine = this.synthesizer.engine || 'standard';
     const salt = cs.callSid;
     const credentials = cs.getSpeechCredentials(vendor, 'tts');
 
-    this.logger.info({language, voice}, `Task:say - using vendor: ${vendor}`);
+    this.logger.info({language,
```
davehorton, line 37: this looks like it was useful for debugging purposes, but its a lot of logging that doesn't seem to add much value at this point.

PR #74, opened 2022-02-26
Feature/gather enhancements alex

akirilyuk:
* start listening after gather say task has finished
* enable/disable listen after play/say child of gather has finished (this allows making gather a bargable say task)


davehorton:
For the moment, I don't believe that an option to "not listen after speech" is good for the gather verb, so for the time being I am not accepting this PR.  Will keep it open for more consideration

PR #111, opened 2022-05-17
issue# 107: on gather timeout, if minDigits are collected, resolve wi…

pgujjeti:
issue# 107: on gather timeout, if minDigits are collected, resolve with 'dtmf-num-digits' instead of 'timeout'


davehorton (changes requested):
looks good, would you mind changing to:

```js
    this._timeoutTimer = setTimeout(() => {
      this.resolve(this.digitBuffer.length >= this.minDigits ? 'dtmf-num-digits' : 'timeout');
    }, this.timeout);

```

davehorton (approved):

PR #132, opened 2022-07-18
Feature/fast http client

davehorton:


davehorton:
This PR replaces the use of [bent](https://www.npmjs.com/package/bent) with [undici](https://www.npmjs.com/package/undici) which offers much better performance calling webhooks

PR #148, opened 2022-08-15
feat: add alert for jambonz parsing falure

xquanluu:


davehorton (changes requested):
These changes look good, but I would like for you to add a new alert type for INVALID_APP_PAYLOAD.  Can you make a PR to this repo:

https://github.com/jambonz/time-series/blob/main/index.js

where you add that new alert type, and you also add call_sid to the fields for alerts?  I will accept that PR, then we can build a new version of that and then update this PR to use it

davehorton (approved):

PR #158, opened 2022-08-24
feat: play verb url support single or array list url

xquanluu:


davehorton (changes requested):
This actually looks good, but in testing I experienced some failures coming from unrelated code, but I wonder if you could fix in this PR.

1) I noticed a bug in conference.js when testing your new play tests.  The problem is that we periodically execute the waitHook, and there is a race condition where we can do so even after the caller has hung up.  The fix is simple [here is the diff](https://gist.github.com/davehorton/d5bb609ce2bef19c680c8de8f92a560d) .

2) In the test case for create-call, we need to give the sipp uas process that is receiving the INVITE some time to come up before we run the test and send the call.  I was getting errors because we were trying to send the INVITE to an address/port that nothing was listening on yet because sipp did not have time to come up.  [The diff is here](https://gist.github.com/davehorton/142197eaf79433e28fd06aacceb6b50d)

davehorton (approved):

PR #174, opened 2022-10-11
Add call_termination_by on app call status

juansalvatella:


davehorton:
closed as this has been separately merged

PR #176, opened 2022-10-19
Fix to handle EADDRINUSE exception when running multiple feature serv…

Megamuch:
…er services on the same machine.

.

### DIFF
--- a/app.js
+++ b/app.js
@@ -15,6 +15,7 @@ const tracer = require('./tracer')(process.env.JAMBONES_OTEL_SERVICE_NAME || 'ja
 const api = require('@opentelemetry/api');
 srf.locals = {...srf.locals, otel: {tracer, api}};
 
+const useMultiple = process.env.JAMBONES_IGNORE_EADDRINUSE || false;
 const PORT = process.env.HTTP_PORT || 3000;
 const opts = {level: process.env.JAMBONES_LOGLEVEL || 'info'};
 const pino = require('pino');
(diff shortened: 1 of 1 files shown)


app.js
```diff
@@ -106,6 +115,7 @@ sessionTracker.on('idle', () => {
 
 const getCount = () => sessionTracker.count;
 const healthCheck = require('@jambonz/http-health-check');
+const { on } = require('events');
```
davehorton, line 118: is this used?

app.js
```diff
@@ -92,7 +93,15 @@ app.use((err, req, res, next) => {
   logger.error(err, 'burped error');
   res.status(err.status || 500).json({msg: err.message});
 });
+
 const httpServer = app.listen(PORT);
+process.on('uncaughtException', function(err) {
+    if(err.errno === 'EADDRINUSE' && useMultiple) {
```
davehorton, line 99: I'm not sure 'useMultiple' is needed.  It seems completely redundant with the information provided the `JAMBONES_IGNORE_EADDRINUSE` env.  Or, to put it another way, if someone configured the system with `JAMBONES_IGNORE_EADDRINUSE` isn't that enough to tell us that this new code needs to be executed?

app.js
```diff
@@ -92,7 +93,15 @@ app.use((err, req, res, next) => {
...
+        const httpServer = app.listen(PORT++);
```
davehorton, line 100: We should burp out a log message here that we are now going to attempt to bind to a new port

davehorton, line 100: It seems to me that this error will also be thrown from lib/utils/aws-sns-lifecycle, yet here we are assuming that the bind error can only come from the healthcheck port?

davehorton (changes requested):
It seems to me that

hkrutzer:
This doesn't really seem like a responsibility for Jambonz to handle. Rather it is the job of configuration management tools. And if it is, a more robust solution like https://github.com/sindresorhus/get-port should probably be used.

app.js
Megamuch: correct. I removed it.

app.js
Megamuch: added.

davehorton:
I'm addressing this in a different PR

PR #183, opened 2022-10-31
Fix/play timeout

two56:
Update timeoutSecs to work outside of FreeSwitch to allow for looping audio.

A scenario where a 15 second audio clip is set to timeout after 20 seconds and loop is > 1, the audio clip will continue to be played and the webhook called after every iteration.

playbackLastOffsetPos is still the last position in the current iteration.
playbackSeconds and playbackMilliseconds are cumulative.
seekOffset is reset after the first iteration.

I've exposed callStatus to take account of the caller hanging up, I wasn't sure if this is the best way of doing it.

The try catch within the timeout is pointless, I added it encase kill was ever to throw.

### DIFF
--- a/lib/tasks/play.js
+++ b/lib/tasks/play.js
@@ -1,5 +1,5 @@
 const Task = require('./task');
-const {TaskName, TaskPreconditions} = require('../utils/constants');
+const {TaskName, TaskPreconditions, CallStatus} = require('../utils/constants');
 
 class TaskPlay extends Task {
   constructor(logger, opts) {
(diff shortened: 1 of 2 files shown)


davehorton:
sorry for the delay in looking at this.  Could you rebase this on main and re-push and then I will test?

lib/tasks/play.js
```diff
@@ -34,14 +48,24 @@ class TaskPlay extends Task {
...
+          if (this.seekOffset >= 0) {
+            file = {file: this.url, seekOffset: this.seekOffset};
+            this.seekOffset = -1;
+          }
           const result = await ep.play(file);
-          await this.performAction(Object.assign(result, {reason: 'playCompleted'}),
-            !(this.parentTask || cs.isConfirmCallSession));
+          playbackSeconds += parseInt(result.playbackSeconds);
+          playbackMilliseconds += parseInt(result.playbackMilliseconds);
+          if (cs.callStatus === CallStatus.Completed || !this.loop || completed) {
```
davehorton, line 59: the proper way to check if the caller hung (or at least to be consistent with other code) is simply to check if the task has been killed, ie `this.killed`.  Because when the caller hangs up we do kill the currently-executing task.  So I would like to see that change (though you can still expose the new CallStatus getter on the CallSession class)

davehorton (changes requested):

lib/tasks/play.js
two56: Change made, sorry I should've spotted that. I was thinking the completed flag might have become unnecessary as the timeout calls `this.kill`, but I still thinks it required for the case where timeout or loop haven't been set.

davehorton (approved):

PR #189, opened 2022-11-11
Update Node Image

guilherme-rauen:
This PR updates the Node image to the latest stable version. And it is the most secure one, with no vulnerabilities on Snyk scan.


davehorton:
@guilherme-rauen you probably want to make a similar PR on the prod/v0.7.7-branch

PR #213, opened 2023-01-09
Fix/conference wait hook

two56:
A small issue we ran into where an old conference wait hook is called even if the wait hook has been updated, causing unexpected audio. I deref the URL when setting a new wait hook so the previous wait hook stops being called.

We also noticed when re-directing a call out of a conference, while a wait hook is in use, the audio wasn't stopping. I've added two `_playSession.kill()` on what I think are the only exit paths. I'm happy to make this a separate PR.

### DIFF
--- a/lib/tasks/conference.js
+++ b/lib/tasks/conference.js
@@ -427,13 +427,19 @@ class Conference extends Task {
         .catch((err) => this.logger.info({err}, 'Error deafing or undeafing participant'));
     }
 
+    if (wait_hook) {
+      if (this.wait_hook)
+        delete this.wait_hook.url;
+      this.wait_hook = {url: wait_hook};
+    }
+
     if (hookOnly && this._playSession) {
       this._playSession.kill();
       this._playSession = null;
     }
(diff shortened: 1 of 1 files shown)


davehorton (approved):

davehorton:
uh....the tests fail now.  I need to change the github actions to run the tests on a PR from now on.  But you should always run them locally yourself before pushing a PR.  To do that, create a file named `run-tests.sh` in the project folder, and add this to it
```
#!/bin/bash
GCP_JSON_KEY='{"type":"service_account","project_id":"drac......etc....your google json key"}' \
JWT_SECRET=foobar \
npm test
```

Since you have broken it (haha) would you mind creating this file and see if you can run the tests locally and figure out what is failing @two56 ?

davehorton:
also always run `npm run jslint` before a PR as well.

davehorton:
actually, the tests are passing locally for me, so let me look into this.

Do get set up so that you can run the tests yourself locally though

two56:
Will do, cheers Dave :)

PR #219, opened 2023-01-17
fix: wrong stt vendor will raise ws notification

xquanluu:


lib/tasks/gather.js
```diff
@@ -162,8 +162,9 @@ class TaskGather extends Task {
         alert_type: AlertType.STT_NOT_PROVISIONED,
         vendor: this.vendor
       }).catch((err) => this.logger.info({err}, 'Error generating alert for no stt'));
-
-      throw new Error(`no speech-to-text service credentials for ${this.vendor} have been configured`);
```
davehorton, line 166: I was thinking we still wanted to throw the error, just that before doing so we should send the jambonz:error message to the listening websocket client.

lib/tasks/gather.js
```diff
@@ -385,7 +386,8 @@ class TaskGather extends Task {
         break;
 
       default:
-        throw new Error(`Invalid vendor ${this.vendor}`);
```
davehorton, line 388: same comment as above

davehorton (changes requested):

lib/tasks/gather.js
xquanluu: fixed

lib/tasks/gather.js
xquanluu: fixed

davehorton (approved):

davehorton:
In the future @xquanluu please put the issue number prefaced by hash in the commit notice, as that will put a nice link in the issue to the PR

PR #225, opened 2023-01-20
fix: split ssml to correct chunks

### LINKED ISSUE #224: Bug: breakLengthyTextIfNeeded not working as expected (generates invalid ssml)
One identified issue regarding ssml text "chunking" which produces an invalid ssml

- A single tag enclosing the whole ssml produces only one chunk with an additional `</speak>`at the end. For example, a lengthy ssml text of the form `<speak><any-tag-name>a very long text</any-tag-name></speak>` results in a single chunk of the form `<speak><any-tag-name>a very long text</any-tag-name></speak></speak>`

Further investigation of this issue led me to discover unexpected behavior of the chunking algorithm.

For example, using the same above example, after finding the `</any-tag-name>` tag closure, the algorithm extracts  `/spea` as the name of the new tag.

From the debug logs:

`breakLengthyTextIfNeeded: searching forward for <//spea>`

I believe that the algorithm should be revisited a bit, or at least introduce a few tests to ensure it works properly.

For the time being, the main issue here is how it behaves when only the text contains only one chunk. The main issue lies here

https://github.com/jambonz/jambonz-feature-server/blob/1b5b37184b3ddd969d422751c3580b32d7e0f46b/lib/tasks/say.js#L80-L91

If there is only one chunk and the startPos is zero, there is no need to append a `</speak>` as the text already ends with one.

So this could be fixed by changing line 84 to

`result.push(0 === startPos ? chunk : `<speak>${chunk}`);`

xquanluu:
fixed #224

### DIFF
--- a/lib/tasks/say.js
+++ b/lib/tasks/say.js
@@ -18,7 +18,7 @@ const breakLengthyTextIfNeeded = (logger,  text) => {
   try {
     return pollySSMLSplit.split(text);
   } catch (err) {
-    this.logger.info({err}, 'Error spliting SSML long text');
+    logger.info({err}, 'Error spliting SSML long text');
     return [text];
   }
 };
--- a/lib/tasks/say.js
+++ b/lib/tasks/say.js
@@ -19,7 +19,7 @@ const breakLengthyTextIfNeeded = (logger,  text) => {
     return pollySSMLSplit.split(text);
   } catch (err) {
     throw new Error(`There is error ${err} while splitting SSML, please check syntax. 
-    if the sentence is too long, please use ,;!? to split the sentence`);
+    if the sentence is too long, please use .,;!? to split the sentence`);
   }
 };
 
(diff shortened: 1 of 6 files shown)


lib/tasks/say.js
```diff
@@ -1,96 +1,26 @@
...
+    // MAX length, exclude 15 characters <speak></speak>
+    hardLimit: chunkSize - 15,
+    // Set of extra split characters (Optional property)
+    extraSplitChars: ',;!?',
+  };
+  pollySSMLSplit.configure(options);
+  try {
+    return pollySSMLSplit.split(text);
+  } catch (err) {
+    throw new Error(`There is error ${err} while splitting SSML, please check syntax.
```
davehorton, line 21: I need more details on in what circumstances we throw this error.  I'm not sure we should ever throw an error, we should handle all valid ssml regardless of length.

davehorton (changes requested):

davehorton (approved):

Issue #229, opened 2023-01-25
createCall: add a default behavior if the trunk isn't defined


vcidst:
With createCall API, the trunks are chosen at random if they're not specified. This causes numbers being dialed out from unrelated trunks. If the trunk isn't specified, check if the number matches any phone numbers already added. If it does, we should use that specific trunk

[link](https://drachtio.slack.com/archives/C017ML321NK/p1674633859665889) to slack discussion

PR #230, opened 2023-01-25
createCall: add a default behavior if the trunk isn't defined

vcidst:
Addresses https://github.com/jambonz/jambonz-feature-server/issues/229

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,4 +89,4 @@ module.exports = {
 
 #### Running the test suite
 
-Please [see this]](./docs/contributing.md#run-the-regression-test-suite).
+Please [see this](./docs/contributing.md#run-the-regression-test-suite).
--- a/lib/http-routes/api/create-call.js
+++ b/lib/http-routes/api/create-call.js
@@ -87,11 +87,11 @@ router.post('/', async(req, res) => {
 
     /**
      * trunk isn't specified,
-     * check if number matches any of our trunks
+     * check if number matches any existing numbers
      *  */
     if (target.type === 'phone' && !target.trunk) {
-      const {lookupTrunk} = dbUtils(this.logger, srf);
-      const voip_carrier_sid = await lookupTrunk(req.body.account_sid, target.number);
+      const {lookupCarrierByPhoneNumber} = dbUtils(this.logger, srf);
+      const voip_carrier_sid = await lookupCarrierByPhoneNumber(req.body.account_sid, target.number);
       logger.info(
         `createCall: selected ${voip_carrier_sid} for requested phone number: ${target.number || 'unspecified'})`);
       if (voip_carrier_sid) {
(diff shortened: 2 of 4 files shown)


davehorton (changes requested):
This seems like it will only work in the case of a REST outdial.  The same feature should work for a dial verb from an inbound call.

vcidst:
I've added these checks also for the dial verb. I'll test it on my cluster, run the regression suite and report back

sammachin:
I don't think this will work in the case of a regex in a number either? only for an exact match? That may be acceptable but it should be documented if that is the case

vcidst:
@sammachin Can you share example of a request with regex number? We can add regex to the query but we'll need to define behaviors for 0, 1 or >1 results. That might add to the complexity

sammachin:
the regex is on the number in the database not the request, its used to match a whole range of incoming numbers to an application, for example you could configure `447.*` to handle all calls to UK 07xxx numbers 
image

vcidst:
Do you have suggestions about what's the expected response when dealing with regex entries in phone numbers column?

davehorton:
@xquanluu could you review this PR?  Check the linked issue at the top of this thread for context

xquanluu:
chhecking PR

xquanluu:
Hi @vcidst @davehorton I have some small question and hope you help me:

1/ jambonz phone number is for defining incoming call phone numbers, why do you use target.number to find the trunk? this can be bad choice as the target.number may be vary and have not defined yet in the number.

2/ I'm thinking to add default trunk sid to account/SP table, this may be better.

vcidst:
>  jambonz phone number is for defining incoming call phone numbers, why do you use target.number to find the trunk?

In India, I cannot use arbitrary carriers to make outbound calls from a number. If the number has been bought from specific carrier, I am allowed to make outbound calls from the number only via the same carrier. For example, if i have a number `1234` from carrier `Airtel` which allows me to make inbound and outbound calls. 
- For inbound calls, I can add `1234` with carrier `Airtel` to Jambonz and connect it to an application. 
- For outbound calls, it needs to be made only via carrier `Airtel` otherwise it fails.

This might be the case for other countries besides India as well. To summarise, this change request does the following,

1. Check if the target.number exists as Jambonz Number
2. If yes, use the same carrier that the number is associated with
3. If not, carry on execution as usual

vcidst:
@davehorton You're correct, I had it mixed up. I've clarified my comments and pushed a commit to update `create-call.js`. I think `dial.js` would have a similar problem too -- not sure which variable to use over there.

xquanluu:
Hi @vcidst 
Now it's good. Could you please add a testcase to make sure the feature is working?

davehorton (approved):

davehorton:
Darn it.  This PR was defective.  After all the discussion above about getting it right in terms of being based on caller id, no one ever fixed and it also got through our testing because we never implemented a test case.  The feature as released does not work and now needs to be fixed.  I have opened [this issue ](https://github.com/jambonz/jambonz-feature-server/issues/359)on it.  Let this be a lesson to us, this was sloppy work.

PR #231, opened 2023-01-25
feat: add dial verb testcases

xquanluu:


davehorton (approved):

davehorton:
tests failed @xquanluu  any idea why?

xquanluu:
Retrigger it's passed. let me triggered one more time

davehorton:
still failing, now on the new dial tests

davehorton:
we cant' have these tests randomly failing or they are of no use.  Why are these listen tests so flaky?

xquanluu:
This is because we are expecting number of packet should be in a range of number. but due to env, for some reason, the total number of receive package is less than expected range.

I'm checking this is issue of env machine is weak or there is some delay by sipp and the loop for playing audio is just 1, I would increase the value to 2 or 3 loop

Issue #232, opened 2023-01-25
No webhook request if user hangs up during early media playback


Megamuch:
I've found that the call status webhook is not called if a caller disconnects during an early media playback call.

The incoming webhook works, but the disconnect hook is not triggered.

davehorton:
can you provide a log from jambonz-feature-server for such a call with loglevel debug?

Megamuch:
`8|jambonz-feature-server  | {"level":20,"time":1674664400333,"pid":203988,"hostname":"ip-10-1-13-62","callId":"d1da1550-1770-123c-71a0-029fb7b14c97","callSid":"e08ca4e8-9eac-4dad-8896-b169f0dffce7","accountSid":"9351f46a-678c-43f5-b8a6-d4eb58d131af","callingNumber":"+XXXXXXXXXX","calledNumber":"+YYYYYYYYY","traceId":"480b530e5e4b8ebf691be2fe9b6db358","url":"https://jbz-media.soundofdata.app:9443/node-red/entertowin","absUrl":"https://jbz-media.soundofdata.app:9443/node-red/entertowin","hdrs":{"Jambonz-Signature":"t=1674664400,v1=add4ecd55affbdf19bad57989c92622499441b95ea849ecfd69787c4dfba1ecf","b3":"480b530e5e4b8ebf691be2fe9b6db358-51ab95312cb3e89b-1","Content-Type":"application/json"},"msg":"send webhook"}
[...]

PR #233, opened 2023-01-26
fix: add early media testcase

xquanluu:


xquanluu:
Hi @davehorton I'm trying to reproduce the #232 but have not success yet. The testcase showing that callStatus still send when caller CANCEL the call while 183 is being processed.

Can you give me some hint if I missed something?

davehorton (approved):

PR #244, opened 2023-02-04
ws-requestor unit test

xquanluu:


lib/utils/ws-requestor.js
```diff
@@ -95,7 +95,7 @@ class WsRequestor extends BaseRequestor {
     if (type === 'session:reconnect') payload = this._sessionData;
     assert.ok(url, 'WsRequestor:request url was not provided');
 
-    const msgid = short.generate();
+    const msgid = this.call_sid;
```
davehorton, line 98: what is the purpose of this change?  I asked for test cases, not a change to functionality

davehorton (changes requested):

lib/utils/ws-requestor.js
xquanluu: There is usecase:

client --connect (data with msgID1)--> wss
client <-- open ---- wss

client <--close -----wss
client  ----reconnect (data with msgID2)--> wss
client <---- message (ACK with msgID2)--- wss

client timeout due to ACK message contains MSGID2 which is not match with msgID1.

I see the the WsRequestor keep generating new msgId everytime it send data to WSS, but the class still expecting the ACK message contains msg of initial request.

davehorton: The reconnect **should** have a new msgid.  Your fix is to re-use the initial msgid again.  I think the proper fix is to properly match to msgid2 in the reconnect

davehorton:
@xquanluu could you continue on with this and improve the test coverage a bit, then comment back in the other test cases?

xquanluu:
Hi @davehorton  thanks for helping me on fixing the reconnect ack message, really nice fix actually.

I have added 2 more testcase for error and unexpected response, and the result is 70% code coverage for the class. I cannot bring it to higher value. May be this is a good start point already for this class, wdyt?

davehorton (approved):

PR #283, opened 2023-03-25
add ENCRYPTION_SECRET variable

avoylenko:
- "ENCRYPTION_SECRET" env added
- fix typo in variable name


davehorton (approved):

lib/utils/encrypt-decrypt.js
```diff
@@ -2,9 +2,9 @@ const crypto = require('crypto');
 const algorithm = process.env.LEGACY_CRYPTO ? 'aes-256-ctr' : 'aes-256-cbc';
 const iv = crypto.randomBytes(16);
 const secretKey = crypto.createHash('sha256')
-  .update(String(process.env.JWT_SECRET))
+  .update(process.env.ENCRYPTION_SECRET || process.env.JWT_SECRET)
```
davehorton, line 5: Please note how this is failing in the github action flow test.  I think you need to restore wrapping it in a String() to make sure it is always a string.  If so, the same applies to the api-server change I think

davehorton (changes requested):

lib/utils/encrypt-decrypt.js
avoylenko: Github action is failing due to env variables. Either ENCRYPTION_SECRET or JWT_SECRET must be supplied.

avoylenko: String(undefined) produces 'undefined' as a string.

davehorton: ok, how do you propose fixing the failing test then?  I am not going to merge until tests pass

avoylenko: @davehorton should be fine now

davehorton (approved):

Issue #201, opened 2022-12-21
feature: ability to queue play requests for the listen verb


vcidst:
Currently, the applications using `listen` verb need to queue play requests and send them upon receiving a `playDone` event from Jambonz. A feature to allow applications to queue audio would be good to have, this queued audio can be played on the call one after another.

It will also be helpful for applications that wish to send the audio to Jambonz as smaller chunks (as opposed to the whole audio file). This might be the case if their audio source is a stream.

Other products of similar nature already handle this feature,
https://www.twilio.com/media-streams
https://support.exotel.com/support/solutions/articles/3000108630-working-with-the-stream-and-voicebot-applet

xquanluu:
Hi @vcidst can you help to give an example for Jambonz Json app for this and what is your expectation for the behavior?
I'm sorry that I cannot catch the idea much.

xquanluu:
closed as PR merged

PR #290, opened 2023-03-30
feat: queue play audio

xquanluu:
#201


davehorton (changes requested):
This looks good.  However, what you have done is to switch the behavior from interrupt current play, to queue play commands.  I actually want this to be an option that is specified when they send us a listen command, and I want the default to be to interrupt.  But on the play command they send they should be able to specify that they want the play command queued behind any that are currently in progress

Issue #252, opened 2023-02-07
opts.auth assigned incorrect value on createCall


ajukes:
When target of create call has an auth object, the call fails with a 500 server error. On investigation it looks like in the following file  `./lib/http-routes/api/create-call.js` 

```
Line 107: if (target.auth) opts.auth = this.target.auth;
```

should be

```
Line 107: if (target.auth) opts.auth = target.auth;
```

PR #296, opened 2023-04-03
Fix incorrect target.auth reference in createCall

ajukes:
API generated calls to registered clients as the initial target fail due to no credentials. This is due to create-call.js referencing `this.target.auth` instead of `target.auth` as its a locally defined constant.


davehorton (approved):

ajukes:
Cant seem to link an issue to a PR

 [Here it is](https://github.com/jambonz/jambonz-feature-server/issues/252)

PR #303, opened 2023-04-05
Support for Cisco NBR for Agentassist

vdharashive:


lib/utils/siprec-utils.js
```diff
@@ -48,8 +48,17 @@ const parseSiprecPayload = (req, logger) => {
     }
 
     if (!sdp || !meta) {
-      logger.info({payload: req.payload}, 'invalid SIPREC payload');
-      return reject(new Error('expected multipart SIPREC body'));
+      if (!meta && sdp) {
```
davehorton, line 51: nesting this in the if statement above makes it difficult to follow.  can you just have if..else if else

lib/session/siprec-call-session.js
```diff
@@ -26,9 +26,9 @@ class SipRecCallSession extends InboundCallSession {
       this.metadata = reMetadata;
 
       if (this.ep && this.ep2) {
-        let remoteSdp = this.sdp1.replace(/sendonly/, 'sendrecv');
+        let remoteSdp = this.sdp1.replace(/sendrecv/, 'sendonly');
```
davehorton, line 29: I think this will actually be a problem.  See the most recent code in this file with the comment
```
      /* note: in order to get the two streams to exchange media through freeswitch, we need them to be sendrecv */
      let remoteSdp = this.sdp1.replace(/sendonly/, 'sendrecv');
```

davehorton (changes requested):

lib/session/siprec-call-session.js
vdharashive: but this ack that needs to be sent to source as sendonly , for the freeswitch is done below

https://github.com/vdharashive/jambonz-feature-server/blob/main/lib/session/siprec-call-session.js#L49

vdharashive: @davehorton you are right , ill revert this commit

lib/utils/siprec-utils.js
```diff
@@ -47,8 +47,16 @@ const parseSiprecPayload = (req, logger) => {
...
 
-    if (!sdp || !meta) {
-      logger.info({payload: req.payload}, 'invalid SIPREC payload');
+
+    if (!meta && sdp) {
+      const arr = /^([^]+)(m=[^]+?)(m=[^]+?)$/.exec(sdp);
+      opts.sdp1 = `${arr[1]}${arr[2]}`;
+      opts.sdp2 = `${arr[1]}${arr[3]}\r\n`;
+      opts.sessionId = uuidv4();
+      logger.info({ payload: req.payload }, 'Cisco NBR payload');
```
davehorton, line 56: minor thing, but Cisco NBR might not be the only siprec payload that lacks metadata.  Can you modify the log statement to say something like "SIPREC payload with no metadata (e.g. Cisco NBR)"

davehorton (approved):

PR #306, opened 2023-04-09
Feature: Use callSid as Jaeger traceId

ajukes:
By luck it appears the jaeger traceId needs to be a valid 16 byte hex. This is easily converted from a v4 UUID and means we can use the callSid as the traceId.

The idea is that we could then implement this for sbc-inbound, sbc-outbound and ultimately trace all the sip services without the need for complex logic.

Confirmed trace in Jaeger UI

image

### DIFF
--- a/lib/utils/call-tracer.js
+++ b/lib/utils/call-tracer.js
@@ -1,5 +1,7 @@
 const {context, trace} = require('@opentelemetry/api');
 const {Dialog} = require('drachtio-srf');
+const {uuid} = require('uuid');
+
 class RootSpan {
   constructor(callType, req) {
     let tracer, callSid, linkedSpanId;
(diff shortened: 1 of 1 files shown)


lib/utils/call-tracer.js
```diff
@@ -9,20 +10,20 @@ class RootSpan {
...
       callSid = dlg.callSid;
       linkedSpanId = dlg.linkedSpanId;
-    }
-    else {
+    } else {
       tracer = req.srf.locals.otel.tracer;
       callSid = req.locals.callSid;
     }
     this._span = tracer.startSpan(callType || 'incoming-call');
+    this._span._spanContext.traceId = this.getHexValue(callSid);
```
davehorton, line 18: do you know if this is a valid way to set a traceId?

PR #311, opened 2023-04-11
Set span context traceId to a hex version of callSid

ajukes:
Create a custom propagator to set the span context with a trace id reflecting a hex version of the callSid.

This should allow us to have uncoupled servers/services under the same traceId in Jaeger.

### DIFF
--- a/lib/utils/call-tracer.js
+++ b/lib/utils/call-tracer.js
@@ -1,39 +1,42 @@
-const {context, trace} = require('@opentelemetry/api');
+
+const {trace, context, SpanKind} = require('@opentelemetry/api');
 const {Dialog} = require('drachtio-srf');
+const {SipPropagator} = require('./sip-propagator');
+
 class RootSpan {
   constructor(callType, req) {
-    let tracer, callSid, linkedSpanId;
+    const tracer = req.srf.locals.otel.tracer;
+    const propagator = new SipPropagator();
+    const ctx = propagator.extract(context.active(), req);
+    this._span = tracer.startSpan(callType || 'incoming-call', {
+      kind: SpanKind.CONSUMER,
+      attributes: this.getSpanAttributes(req),
+      root: false
+    }, ctx);
+    this._ctx = trace.setSpan(ctx, this._span);
+    this.tracer = tracer;
+  }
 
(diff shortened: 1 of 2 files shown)


davehorton:
also can you confirm this works with inbound calls as well as REST api triggered outdials?

ajukes:
SipPropagator  works similar to B3Propagator. [here](https://open-telemetry.github.io/opentelemetry-js/classes/_opentelemetry_propagator_b3.B3Propagator.htm)

Using the extract method to return a Context which then is used to create the span. I mean I could simply do the following in the call-tracer.js file

```
const ctx = trace.setSpanContext(context, {
      traceId,
      spanId,
      isRemote: true,
      traceFlags: TraceFlags.SAMPLED
    });
```

But it seemed cleaner to do the extracting of traceId, spanId in the SipPropagator  class. I haven't hooked this into the tracer config but I suppose that could be the next natural step. Then instead of manually instantiating SipPropagator we could just leverage `propagation.extract(context.active(), input);` seen [here ](https://opentelemetry.io/docs/instrumentation/js/propagation/)

What are your thoughts? Unless I have completely missed the mark here.

I haven't tested REST calls yet, I will update the PR once tested.

ajukes:
I can see REST call in jaeger with a traceID representing the callSid so the propagator works as expected. However, what jambonz saves to the DB is "000000000000.."

image

davehorton:
can you set this PR to draft status for now?

davehorton:
@xquanluu please give your review

xquanluu (approved):
It's good, it's really nice work

ajukes:
thanks @xquanluu, I have made a very small change. `traceId` now is determined from `callId` not `callSid` which allows us to attach spans much earlier on in apps such as sbc-inbound.

ajukes:
Hi, I believe this is now ready for final review and please checkout the following PRs to have end-end tracing:

- https://github.com/jambonz/jambonz-infrastructure/pull/44
- https://github.com/jambonz/sbc-outbound/pull/79
- https://github.com/jambonz/sbc-inbound/pull/96

davehorton:
Hi @ajukes could you rebase to latest main and push?

PR #313, opened 2023-04-12
Bug/fix env in tracer

Catharsis68:
Fix usage of `OTEL_EXPORTER_COLLECTOR_URL` in tracer.js


davehorton:
fixed in separate commit

PR #320, opened 2023-04-18
Bug/gather race condition

Catharsis68:
This PR fixes a bug  where two `gather` could be active at the same time since we did not await the kill of the first one before creating the second one.


davehorton:
THanks, merged similar fix in https://github.com/jambonz/jambonz-feature-server/pull/321 instead.

Issue #317, opened 2023-04-15
awselb needs user-agent header mandatory for http and wss


vdharashive:
6|jambonz-feature-server  | {"level":30,"time":1681490966377,"pid":878525,"hostname":"ip-172-31-11-177","callId":"284cdc42-5587-123c-75a7-0e9f95109ce9","callSid":"603fc6f5-8fe5-4efe-9aab-79f109165deb","accountSid":"3978e7ef-e014-4a9e-9184-ff24a19d9fad","callingNumber":"+646454645","calledNumber":"+45345354","traceId":"9dab3e969562bd488c82990881026331","headers":{"server":"awselb/2.0","date":"Fri, 14 Apr 2023 16:49:26 GMT","content-type":"text/html","content-length":"118","connection":"keep-alive"},"statusCode":403,"statusMessage":"Forbidden","msg":"WsRequestor - unexpected response"}

davehorton:
please provide more details when opening an issue.  Not really clear on this, not enough info to start working on it.

vdharashive:
Ill fix this issue  and Raise PR 

reference 
https://repost.aws/knowledge-center/waf-block-http-requests-no-user-agent

davehorton:
can you make PR?

PR #327, opened 2023-04-27
awselb needs user-agent header mandatory for http and wss #317

vdharashive:


davehorton:
@xquanluu  can you please give a review?

lib/utils/http-requestor.js
```diff
@@ -91,7 +91,7 @@ class HttpRequestor extends BaseRequestor {
     const url = hook.url || hook;
     const method = hook.method || 'POST';
     let buf = '';
-
+    httpHeaders = {['user-agent']: 'jambonz'};
```
davehorton, line 94: I think this should be controlled by the presence of an env var that optionally provides http headers to be used on callbacks in the form of a json-ified string

vdharashive: JAMBONZ_WEB_HOOK_HEADER = {['user-agent']: 'jambonz'} something like this ill have it done

davehorton: just checking if you still are planning on this

davehorton:
I believe this was merged in a separate PR

Issue #333, opened 2023-05-05
Bug: REST update call timing issue


two56:
Reproduce issue:

1. Place a REST call to update the `call_hook` while an existing `actionHook` is in-flight.
2. Place a REST call to update the `call_hook` just before an existing `actionHook` is about to take flight but the updated `call_hook` hasn't responded.

Behaviour:

1. There's a race condition in both cases giving unpredictable results.

Expected behaviour:

The updated `call_hook` overrides any in-flight or about to take flight webhooks.

davehorton:
Thanks, I believe you were going to make a PR for this @two56 ?

PR #334, opened 2023-05-05
Fix: Addressing #333 using a redirect task

two56:
I've swapped out the webhook requests with redirect tasks allowing the currently running application to be replaced immediately. This stops the race condition where both the currently running tasks webhook and the newly requested webhook could run in parallel.

Additionally, I've added an abortController to kill any in-flight request of the task being replaced.

### DIFF
--- a/lib/session/call-session.js
+++ b/lib/session/call-session.js
@@ -998,30 +998,19 @@ ${credential.label ? `, label: ${credential.label}` : ''} is chosen`);
    * @param {object} [opts.call_hook] - new call_status_hook
    */
   async _lccCallHook(opts) {
-    const webhooks = [];
     let sd, tasks, childTasks;
-    const b3 = this.b3;
-    const httpHeaders = b3 && {b3};
-
     if (opts.call_hook || opts.child_call_hook) {
       if (opts.call_hook) {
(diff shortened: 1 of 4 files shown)


davehorton:
This is a proposed fix for #333 I believe?

@xquanluu can you review this?  I want to make sure that we are finding the simplest possible way to address this issue..

davehorton:
@two56 is this fix still valid?  If so can you rebase?  @xquanluu can you review if so

xquanluu:
Hi @two56 I want to reproduce this issue locally to make me understand the problem and the PR, it's great if you can give me the steps.

two56:
Hopefully #333 should give you some information.

In short, there's a race condition when a verb is completing/requesting the next action and an update call REST request is placed at the same time.

The exaggerated version of the scenario is below:

- Allow a verb to reach the actionHook
- When the actionHook is called, trigger the update call REST request and delay the response to the actionHook by 1000-2000ms
- Respond to the update call REST request webhook immediately
- Followed by the delayed actionHook response

What you should see is the REST update verb start to be actioned, shortly followed by the actionHook verb.

The expected behaviour, from my point of view, would be that only the REST update verb is actioned and the inflight actionHook is cancelled.

davehorton:
@xquanluu can you review this again when you have time?

davehorton:
@xquanluu can you test this?

PR #339, opened 2023-05-07
feat: rest:dial amd

xquanluu:


davehorton (commented):
Is it going to be possible to add a test case for this?

davehorton (commented):
Is it going to be possible to add a test case for this?

xquanluu:
Yes, I'm adding the testcase, I'm first check if my changes break legacy testsuite.

davehorton (approved):

davehorton:
seems like test is failing
```
# test create-call amd
Unhandled Rejection at: Promise Promise {
  <rejected> TypeError: Cannot destructure property 'vendor' of 'ep.amd' as it is null.
```

Issue #343, opened 2023-05-09
Bug: REST dial timeout being ignored


two56:
REST dial timeout is ignored and defaults to 60 seconds.

PR #344, opened 2023-05-09
Fix: REST dial timeout

two56:
This fixes a bug I introduced in #268 and addresses #343. When the call session isn't available, use the saved request to cancel.

I still need to check the handling of a 302, in the case of using the saved request to cancel.

### DIFF
--- a/lib/tasks/rest_dial.js
+++ b/lib/tasks/rest_dial.js
@@ -28,6 +28,7 @@ class TaskRestDial extends Task {
   */
   async exec(cs) {
     await super.exec(cs);
+    this.req = cs.req;
     this.canCancel = true;
 
     this._setCallTimer();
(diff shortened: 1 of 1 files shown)


davehorton:
@xquanluu can you review

lib/tasks/rest_dial.js
```diff
@@ -28,6 +28,7 @@ class TaskRestDial extends Task {
   */
   async exec(cs) {
     await super.exec(cs);
+    this.req = cs.req;
```
davehorton, line 31: where is `this.req` actually referenced?

two56: In the kill method. It attempts to use the passed in call session, if that isn't available the saved request is used.

two56:
If a call has been 302'd and the call session isn't passed into the `kill()` method but the saved request is used instead, the call isn't cancelled.

davehorton:
can you rebase to main?

davehorton:
also if you feel it is ready for review please change it from draft status

davehorton:
@two56 if you can get this done in the next hour it can go into the 0.8.3 release otherwise it will wait till 0.8.4

two56:
@davehorton My latest commit isn't being pulled through. Give me a few minutes to try and sort it out.

Issue #353, opened 2023-05-18
app_json is not applied for dial outbound


xquanluu:
When define new jambonz application via postal, I have provided app_json.

this is working well with inbound call, but outbound call, jambonz still query to call hook.

PR #358, opened 2023-05-25
fix: app_json is used for outbound call

xquanluu:
#353

### DIFF
--- a/lib/http-routes/api/create-call.js
+++ b/lib/http-routes/api/create-call.js
@@ -19,7 +19,12 @@ router.post('/', async(req, res) => {
   logger.debug({body: req.body}, 'got createCall request');
   try {
     let uri, cs, to;
+    // app_json is creaeted by only api-server.
+    // if it available, take it and delete before creating task
+    const app_json = req.body.app_json;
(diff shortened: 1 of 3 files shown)


davehorton:
@xquanluu please address test failures

xquanluu:
Hi @davehorton , need your suggestion on app_json for outbound call.

Currently api-server query application by application_sid and get app_json from the application.
This app_json will be added to request sent to feature server for initiating outbound call.
Feature server will do verb validation for rest_dial and see that app_json is unknown param and reject the call.
Now if we modify verb-specification to have app_json for me is bad choice as we allow app_json to be sent from user, which is not correct.
I created a commit do a little bit hack, I'm not like it but I don't have any better idea.
Now in rest:dial task, I create a setter for app_json, create_call rest api will take the app_json from request body and delete it from request body. so we don't break the validation for verb. After that use setter to set app_json to rest:dial object.
https://github.com/jambonz/jambonz-feature-server/pull/358/commits/3ad20bd68c794273e6ee56937faab163455c7317

davehorton (approved):

PR #390, opened 2023-07-07
fix background listen send customerData to api server

xquanluu:


xquanluu:
waiting for user to test it.

PR #391, opened 2023-07-10
Fix/background listen tag

xquanluu:
Fix Feature server don't send customerData to api-server for record all call feature


javibookline:
Hi @xquanluu have you been able to review the comment I added??

That the code in call-session.js

```
// Remove unneeded customer data to be sent to api server.
this.backgroundGatherTask.ignoreCustomerData = true;
```

should be

```
// Remove unneeded customer data to be sent to api server.
this.backgroundListenTask.ignoreCustomerData = true;
```

xquanluu:
Thanks a lot @javibookline to point that out.
I have fixed the issue by your comment, now all testcases are passed.

davehorton (approved):

PR #400, opened 2023-07-20
feat siprec custom header

xquanluu:


davehorton:
update verb specs now

lib/session/call-session.js
```diff
@@ -423,7 +423,10 @@ class CallSession extends Emitter {
           'X-Call-Sid': this.callSid,
           'X-Account-Sid': this.accountSid,
           'X-Application-Sid': this.applicationSid,
-        }
+          ...(this.recordOptions.headers && {'Content-Type': 'application/json'})
```
davehorton, line 426: I'm unclear on this.  Are you saying that due to this change the INFO request now has a body (it didn't before did it?) and this is somehow dependent on their being custom headers?  I'm missing something I think

davehorton (approved):

Issue #403, opened 2023-07-25
Intermittent Client already destroyed on Rest-Dial webhooks


ajukes:
When creating a call from API,  on the FS, 60% of the calls fail to send webhook request/events as http client is already destroyed.

Non working example: 

(from ,to and domain have been redacted)

```
[11:07:56.215] INFO (1331): SessionTracker:add callSid e4e838e3-7daf-4059-8a21-91717735c36f, we have 15 session being tracked
[11:07:56.215] INFO (1331): CallSession:exec starting 1 tasks
    callSid: "e4e838e3-7daf-4059-8a21-91717735c36f"
    callId: "595bd894-a57e-123c-c284-060c0ae6881d"
    accountSid: "0f459985-82a1-4e49-b440-bbe459f1576b"
    traceId: "527d8250b250340026cc9b8de7c0e2e6"
    tasks: "[rest:dial]"
[11:07:56.215] INFO (1331): CallSession:exec starting task #0:1: rest:dial
    callSid: "e4e838e3-7daf-4059-8a21-91717735c36f"
    callId: "595bd894-a57e-123c-c284-060c0ae6881d"
    accountSid: "0f459985-82a1-4e49-b440-bbe459f1576b"
    traceId: "527d8250b250340026cc9b8de7c0e2e6"
[11:07:56.216] INFO (1331): outbound REST call attempt to {"type":"sip","sipUri":"sip:+447777777777@1.1.1.1:5060","vmail":false,"headers":{"P-Asserted-Identity":"\"+441234567890\" <sip:+441234567890@localhost>"}} has been sent
    callSid: "e4e838e3-7daf-4059-8a21-91717735c36f"
    callId: "595bd894-a57e-123c-c284-060c0ae6881d"
    accountSid: "0f459985-82a1-4e49-b440-bbe459f1576b"
    traceId: "527d8250b250340026cc9b8de7c0e2e6"
    sid: "e4e838e3-7daf-4059-8a21-91717735c36f"
[11:08:05.217] INFO (1331): CallSession:exec completed task #0:1: rest:dial
    callSid: "e4e838e3-7daf-4059-8a21-91717735c36f"
    callId: "595bd894-a57e-123c-c284-060c0ae6881d"
    accountSid: "0f459985-82a1-4e49-b440-bbe459f1576b"
    traceId: "527d8250b250340026cc9b8de7c0e2e6"
[11:08:05.217] INFO (1331): CallSession:exec all tasks complete
[...]

PR #405, opened 2023-07-26
Fix/rest dial timeout does not send status callback

xquanluu:
#403

### DIFF
--- a/lib/session/call-session.js
+++ b/lib/session/call-session.js
@@ -796,7 +796,9 @@ class CallSession extends Emitter {
         }
       }
 
-      if (0 === this.tasks.length && this.requestor instanceof WsRequestor && !this.callGone) {
+      if (0 === this.tasks.length && ((this.requestor instanceof WsRequestor && !this.callGone) ||
+        // Tasks are all done, but dialog is not establish yet, waiting for cancel/handup the call.
+        !this.dlg)) {
         //let span;
         try {
           //const {span} = this.rootSpan.startChildSpan('waiting for commands');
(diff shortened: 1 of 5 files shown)


lib/session/call-session.js
```diff
@@ -1790,6 +1792,17 @@ class CallSession extends Emitter {
...
       .catch((err) => this.logger.error(err, 'redis error'));
+
+    if (this.wakeupResolver && !this.dlg && this.isCurrentCallHigherThan(CallStatus.Failed)) {
+      // Someone is waiting for call to be final
+      this.wakeupResolver({reason: 'session ended'});
+      this.wakeupResolver = null;
+    }
+  }
+
+  isCurrentCallHigherThan(target) {
```
davehorton: Wow, I don't understand what this code is doing.  Can you explain ?

lib/session/call-session.js
```diff
@@ -796,7 +796,9 @@ class CallSession extends Emitter {
         }
       }
 
-      if (0 === this.tasks.length && this.requestor instanceof WsRequestor && !this.callGone) {
+      if (0 === this.tasks.length && ((this.requestor instanceof WsRequestor && !this.callGone) ||
+        // Tasks are all done, but dialog is not establish yet, waiting for cancel/handup the call.
+        !this.dlg)) {
```
davehorton, line 801: Are you sure that this won't hit cases other than the outdial failure?  What about an incoming call that an app simply uses a sip:decline verb on?  In that case when the app is over there has never been a dialog.  Can you make sure that case is handled appropriately?

lib/session/call-session.js
davehorton: what does "higher than" mean in this case?

xquanluu: For example: Call is final when callStatus is in [Faled, Busy, NoAnswer, Completed]

Here the function is has typo issue that I change name to isCurrentCallStatusHigherThan

I want to use index of each key,value pair in json to identify if call is already final

lib/session/call-session.js
xquanluu: you are right, add fix to outbound call only.

xquanluu:
sure, Let me think on another approach actually

Issue #406, opened 2023-07-27
tag keys being transformed to snake case


davehorton:
When doing a new outgoing call via REST and using the property "tag" 
the keys are being transformed. 

e.g CamelCase to snake_case and lowercase etc. The tags should remain exactly the same

example REST createCall
```
POST https://xxxx/api/v1/Accounts/:accountSid/Calls
{
    "application_sid": "d6a219b6-683a-4e04-835a-dabd092edf13",
    "from": "441772503124", 
    "to": {
        "type": "phone",
        "number": "4917623722536" 
    },
    
    "tag": {
        "callCount": 10,
        "env": "DEVELOPMENT",
        "processId": "processId",
        "Customer": "Customer",
        "Vehicle": "Vehicle",
        "Event_Camel": "Event_Camel",
        "CamelCase": "CamelCase"
    }
}
```
Results in a webhook payload
```
{
    "userId": "441772503124",
    "data": {
        "payload": {
            "direction": "outbound",
            "trace_id": "bedd0a4a58bd8043ade94697acbc38b3",
            "caller_name": "",
            "call_sid": "5b422af2-f051-4961-8e4d-b71a16565144",
            "account_sid": "b507f72e-be93-4802-a15d-b6286d90e32d",
            "application_sid": "d6a219b6-683a-4e04-835a-dabd092edf13",
            "call_status": "in-progress",
            "call_id": "9ab8ea97-8ae9-123c-a5a6-16826c641508",
            "sip_status": 200,
            "sip_reason": "OK",
            "from": "441772503124",
            "to": "4917623722536",
            "customerdata": {
                "call_count": 10,
                "env": "DEVELOPMENT",
                "process_id": "processId",
                "customer": "Customer",
                "vehicle": "Vehicle",
                "event_camel": "Event_Camel",
                "camel_case": "CamelCase"
            },
            "local_sip_address": "10.1.200.146:5060",
            "defaults": {
                "synthesizer": {
                    "vendor": "microsoft",
                    "language": "en-GB",
                    "voice": "en-GB-LibbyNeural"
                },
                "recognizer": {
                    "vendor": "microsoft",
                    "language": "en-GB"
                }
            }
        },
        "event": "CALL_CREATED"
    },
    "traceId": "bedd0a4a58bd8043ade94697acbc38b3"
}
```

davehorton:
fixed

PR #407, opened 2023-07-27
fix customerdata should not be snakeCase

xquanluu:
#406


lib/utils/http-requestor.js
```diff
@@ -87,7 +87,7 @@ class HttpRequestor extends BaseRequestor {
 
     assert(HookMsgTypes.includes(type));
 
-    const payload = params ? snakeCaseKeys(params, ['customerData', 'sip']) : null;
+    const payload = params ? snakeCaseKeys(params, ['sip']) : null;
```
davehorton, line 90: I may be wrong, but I think this change is an error.  The removed code says **not** to modify (by snake-casing) the keys inside customerData.  Which is correct.  Your change, it would seem, leads to exactly the bug we are trying to fix.

Have you verified this bug in your own testing, by sending a createCall POST

lib/session/call-info.js
```diff
@@ -144,7 +144,7 @@ class CallInfo {
     if (typeof this.duration === 'number') obj.duration = this.duration;
 
     if (this._customerData) {
-      Object.assign(obj, {customerData: this._customerData});
+      Object.assign(obj, {customerdata: this._customerData});
```
davehorton, line 147: I don't understand this.  There are many places in the code where we are referencing 'customerData', none where we are referencing 'customerdata'.  Example, in call-info.js
```
      if (tag) this._customerData = tag;
```
Its unclear to me why this changes fixes the bug, and why it may not be creating other bugs given that we are now referring to the collection of customer tagged data in two different ways in the code

Issue #424, opened 2023-08-13
Support multiple credentials from same speech vendor


davehorton:
Related: https://github.com/jambonz/jambonz-webapp/issues/304

davehorton:
merged

PR #426, opened 2023-08-15
feat: multi speech credential diff labels but same vendor

xquanluu:
#424

### DIFF
--- a/lib/session/call-session.js
+++ b/lib/session/call-session.js
@@ -179,6 +179,16 @@ class CallSession extends Emitter {
   set speechSynthesisVendor(vendor) {
     this.application.speech_synthesis_vendor = vendor;
   }
+
+    /**
+   * default label to use for speech synthesis if not provided in the app
+   */
+    get speechSynthesisLabel() {
+      return this.application.speech_synthesis_label;
+    }
+    set speechSynthesisLabel(label) {
+      this.application.speech_synthesis_label = label;
+    }
   /**
    * default voice to use for speech synthesis if not provided in the app
    */
@@ -206,6 +216,15 @@ class CallSession extends Emitter {
   }
   set speechRecognizerVendor(vendor) {
     this.application.speech_recognizer_vendor = vendor;
+  }
+  /**
+   * default vendor to use for speech recognition if not provided in the app
+   */
+  get speechRecognizerLabel() {
+    return this.application.speech_recognizer_label;
+  }
+  set speechRecognizerLabel(label) {
+    this.application.speech_recognizer_label = label;
   }
   /**
  * default language to use for speech recognition if not provided in the app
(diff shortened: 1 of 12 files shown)


lib/session/call-session.js
```diff
@@ -640,14 +659,16 @@ class CallSession extends Emitter {
...
     const {writeAlerts, AlertType} = this.srf.locals;
     if (this.accountInfo.speech && this.accountInfo.speech.length > 0) {
-      const credential = this.accountInfo.speech.find((s) => s.vendor === vendor);
+      const credential = this.accountInfo.speech.find((s) => s.vendor === vendor &&
+        ((label && s.label === label) || label === null));
       if (credential && (
         (type === 'tts' && credential.use_for_tts) ||
         (type === 'stt' && credential.use_for_stt)
       )) {
+        this.logger.info(`Speech credential vendor: ${credential.vendor} and Label: ${credential.label} is chosen`);
```
davehorton, line 671: I'd prefer the info message not to mention label if label is null

lib/tasks/config.js
```diff
@@ -105,6 +105,9 @@ class TaskConfig extends Task {
       cs.speechSynthesisVendor = this.synthesizer.vendor !== 'default'
         ? this.synthesizer.vendor
         : cs.speechSynthesisVendor;
+      cs.speechSynthesisLabel = this.synthesizer.label !== 'default'
```
davehorton, line 108: why would you compare label to 'default' ?  That doesn't make sense, there is no default label

lib/tasks/config.js
```diff
@@ -117,6 +120,9 @@ class TaskConfig extends Task {
       cs.speechRecognizerVendor = this.recognizer.vendor !== 'default'
         ? this.recognizer.vendor
         : cs.speechRecognizerVendor;
+      cs.speechRecognizerLabel = this.recognizer.label !== 'default'
+        ? this.recognizer.label
```
davehorton, line 124: same comment about default

davehorton (changes requested):

lib/session/call-session.js
xquanluu: fixed

lib/tasks/config.js
xquanluu: I am having same question like you, for example vendor and language are having same behavior. why is it like that?

I think for config verb, when a parameter = default means, the configure verb will use the value provided by application and not overide the parameter value in that call.

lib/tasks/config.js
xquanluu: same as above answer

lib/tasks/config.js
davehorton: the reason for vendor and language is that after setting applications defaults, someone might want to change something like hints with config or say but not vendor or language.  I suppose that you are right....the same could be true of label.  So I will close this

davehorton (approved):

Issue #425, opened 2023-08-14
Feat: Secondary/Failover Speech Provider Option


guilherme-rauen:
Feature Request

We propose the addition of a configuration option for a secondary speech provider within the web application. This feature aims to provide a seamless failover mechanism for both speech-to-text (STT) and text-to-speech (TTS) services. In case, a secondary speech provider is configured in the WebApp, we can use it if an error occurs during the processing of the task using the default one.

The proposed solution involves adding a configuration layer at the application level, allowing users to define not only the default speech provider, but will add maybe a check box that will enable users to define a failover speech provider with the same components used by the default one.

In case of TTS, if the original text has SSMLs, since they can differ from provider to provider, we thought about adding an additional property in the payload for those cases, where we send the `failoverText` just with the plain text.

If I am not mistaken, looks like the changes will be on: the verb definition, feature-server and webapp.

Please share your thoughts and let's discuss your concerns, or additional requirements related to this proposal.

davehorton:
I think we need clarification on some aspects of this feature.  Let's say there is a TTS failure -- authentication fails for some reason -- and the user has configured the application with a backup TTS provider.  So we react by using the backup TTS vendor for that call.  

Now, when the next call for that application comes in, which TTS provider do we attempt?  Do we attempt the primary again?  Or, given that it failed before do we immediately go to the backup?  And it the answer is we go to the backup, when / what would cause us to ever start trying the primary again?

dilekarzu:
We should always try the primary one. We could go one step further and even implement logic rules where if we see three failures we switch to the backup one for the next x hours, but this might just overcomplicate things.

davehorton:
Suggested wireframe

![image](https://github.com/jambonz/jambonz-feature-server/assets/1884973/681e3351-a659-4672-aef8-eb0e6a1f8709)

davehorton:
Before doing this feature, we should complete the [feature request to enable multiple credentials](https://github.com/jambonz/jambonz-webapp/issues/304) for the same speech vendor, because a common use case for this feature would be to have for instance a primary and backup TTS be the same vendor, but in different regions; e.g. primary is Azure in eastus and backup is Azure in eastus2.

davehorton:
Note: this feature request should really have been opened in the jambonz-webapp repo.  Leaving it here for now.

dilekarzu:
[...]

PR #427, opened 2023-08-17
feat fallback speech

xquanluu:
#425


davehorton (changes requested):
update verb-specifications

davehorton:
still needs support for handling gather e.g. STT failures

PR #431, opened 2023-08-18
Fix: REST dial timeout status errors

two56:
Hopefully addressing one of the triggers in #403

Cancel the transaction rather than killing the task so the status and task destruction happen in the normal flow. This allows the client to return a 487, the status update to happen and the call session to finish and clean up.


davehorton:
@xquanluu can you review, and test yourself.  Do we need to add a test case for this?

xquanluu:
Hi @two56 Thank you so much for your PR.

@davehorton I confirmed this PR fixed the issue. I have manually tested on jambonz.one.
For adding testcase, yes we can do it, there is legacy testcase which we can reuse for this. we just need to add code for checking call-status request is sent from jabonz and validate the body.

https://github.com/jambonz/jambonz-feature-server/blob/main/test/create-call-test.js#L22

davehorton (approved):

PR #432, opened 2023-08-19
fix participant join options

avoylenko:
This PR fixes participant join options  and adds a new options to set end conference duration. 

must be approved after https://github.com/jambonz/verb-specifications/pull/21


davehorton (approved):

lib/tasks/conference.js
```diff
@@ -48,7 +48,7 @@ class Conference extends Task {
     this.confName = this.data.name;
     [
       'beep', 'startConferenceOnEnter', 'endConferenceOnExit', 'joinMuted',
-      'maxParticipants', 'waitHook', 'statusHook', 'endHook', 'enterHook'
+      'maxParticipants', 'waitHook', 'statusHook', 'endHook', 'enterHook', 'endConferenceDuration'
```
davehorton, line 51: actually, you can't just add properties to the verb specs like this.  You need to add them to [verb-specifications](https://github.com/jambonz/verb-specifications/blob/3d4df422e1f79360df6c88cc7fc3d9d84fcd058e/specs.json#L177) first, have that published and then update feature-server to that version.  You then also need to update the other apps like api-server that uses verb-specifications

davehorton (changes requested):

lib/tasks/conference.js
davehorton: ah ok sorry, seeing the other PR now

avoylenko:
verb specifications updated to the latest

davehorton (approved):

PR #453, opened 2023-09-06
fix validation for alternative languages

paulotelles:
I noticed some errors when setting an empty array of alternative languages, we have an error on concat function since the altLanguages can be undefined and the concat function does not work. This code will prevent this kind of error to happen, cause it was leading to miss a gather.


paulotelles:
related to #455

davehorton (changes requested):

PR #454, opened 2023-09-06
issue #456 error on validating hints using microsoft

paulotelles:
We can have an error when using Microsoft, since we're not validating if the length "rOpts.hints.length" since if we have no hints there is no lenght, so I added a check for it "rOpts.hints && rOpts.hints.length > 0" instead of "&& rOpts.hints.length > 0". Also I put it inside a try catch block, since without the try catch block we cannot see this kind of error.

### DIFF
--- a/lib/utils/transcription-utils.js
+++ b/lib/utils/transcription-utils.js
@@ -610,7 +610,7 @@ module.exports = (logger) => {
       });
       return opts;
     } catch (error) {
-      logger.error(error, `Error on setting channel vars.`)
+      logger.error(error, 'Error on setting channel vars.')
     }
   };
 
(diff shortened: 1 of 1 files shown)


davehorton:
There are enormous changes here due to the different settings you have for indenting.  You would need to adjust your editor settings to conform to ours (tab size 2, insert spaces when pressing Tab) in order to have PRs accepted.

paulotelles:
I just put the code of this function inside a try catch block, that's why it looks like I changed the whole code of this function.

davehorton:
rather than surround this code with try-catch I would prefer to make sure that no errors are thrown from this code.  This particular code should never throw an error - and I am not currently aware of any scenario where it does.  If you can supply a scenario or some logs showing an error thrown from this code then we can look at it, but the preferred fix would be to ensure this code block never throws, because this code has no dependencies on any external service and therefore it should be written to always work.

paulotelles:
I removed the try catch block so now it's easier to see what are the changes, it's just a validation for the hints on microsoft, cause we're trying to access the property length without validating if the "rOpts.hints" exists first

davehorton (approved):

Issue #448, opened 2023-08-31
feat: Fast speech-to-text recognition


vdharashive:
For some languages, the speech-to-text engine requires a longer time to decide on 'end of single utterance' (this can result in a poor user experience).

If the fast speech-to-text recognition workaround is enabled, will trigger a recognition event after a specified time after the last hypothesis.

https://techdocs.audiocodes.com/voice-ai-connect/#VAIG_Combined/speech_customization.htm#Fast20speech-to-text20recognition

davehorton:
@xquanluu please start with a proposal for the changes to the verb-specifications

davehorton:
merged

PR #461, opened 2023-09-11
feat fast recognition

xquanluu:
#448


davehorton (commented):
This all seems very complicated.  Why do we need to maintain a new buffer `_fastRecognitionTranscripts` with special logic for compiling them?  Isnt this a simple matter of returning the last interim transcript under certain conditions?

xquanluu:
fixed comment

davehorton (approved):

PR #467, opened 2023-09-20
dual streams

xquanluu:


lib/http-routes/api/create-call.js
```diff
@@ -127,16 +127,41 @@ router.post('/', async(req, res) => {
 
     /* launch outdial */
     let sdp, sipLogger;
+    let dualEp;
+    let localSdp = ep.local.sdp;
+
+    if (req.body.dual_streams) {
+      dualEp = await ms.createEndpoint();
+      const sdpParts = dualEp.local.sdp.split('m=');
```
davehorton, line 135: I would prefer if we use the sdp-transform package for manipulating SDPs. we are already using it in utils/siprec-utils.js as an example

davehorton (changes requested):

lib/http-routes/api/create-call.js
xquanluu: start using sdp-transform

davehorton (approved):

PR #474, opened 2023-10-04
add sentenialPassword option

xquanluu:
https://github.com/jambonz/realtimedb-helpers/issues/49


davehorton (approved):

MrFabio:
Is the `SENTINAL` part on purpose due to already having SENTINEL on the env name?
In `JAMBONES_REDIS_SENTINEL_SENTINAL_PASSWORD`

davehorton:
yeah `JAMBONES_REDIS_SENTINEL_SENTINAL_PASSWORD` is ugly.  @xquanluu can you fix (everywhere).  This should be `JAMBONES_REDIS_SENTINEL_PASSWORD` unless I am missing something.

MrFabio:
> yeah `JAMBONES_REDIS_SENTINEL_SENTINAL_PASSWORD` is ugly. @xquanluu can you fix (everywhere). This should be `JAMBONES_REDIS_SENTINEL_PASSWORD` unless I am missing something.

Unfortunately, that env already [exists](https://github.com/jambonz/jambonz-smpp-esme/blob/e400278514accfa6fe082a47be6e73ce90c3f052/app.js#L22C1-L29) for the redis password:
``` js
  ...(process.env.JAMBONES_REDIS_SENTINEL_PASSWORD && {
    password: process.env.JAMBONES_REDIS_SENTINEL_PASSWORD
  }),
  ...(process.env.JAMBONES_REDIS_SENTINEL_SENTINAL_PASSWORD && {
    sentinelPassword: process.env.JAMBONES_REDIS_SENTINEL_SENTINAL_PASSWORD
  }),
```

The correct way would be to use the password for redis and sentinel:
``` js
  ...(process.env.JAMBONES_REDIS_PASSWORD && {
    password: process.env.JAMBONES_REDIS_PASSWORD
  }),
  ...(process.env.JAMBONES_REDIS_SENTINEL_PASSWORD && {
    sentinelPassword: process.env.JAMBONES_REDIS_SENTINEL_PASSWORD
  }),
```
but this would break the current deployments if someone is already using this variable.

xquanluu:
Yes, I had the same but we need to differentiate between   JAMBONES_REDIS_SENTINEL_PASSWORD and JAMBONES_REDIS_SENTINEL_SENTINAL_PASSWORD

because I want to:
1/ prefix: JAMBONES_REDIS_SENTINEL is used for all parameter for sentinal
2/ prefix JAMBONES_REDIS is used for all parameter for standalone redis

Currently I don't see any better, maybe you can suggest diffrent name

MrFabio:
If we can't rename the redis env, we can use something like:
Redis:
`JAMBONES_REDIS_SENTINEL_PASSWORD` - it already exists

Sentinel:
`JAMBONES_REDIS_SENTINEL_CLUSTER_PASSWORD`
or
`JAMBONES_REDIS_SENTINEL_SERVER_PASSWORD`

my only concern with the current naming is the typo _SENTIN**A**L_

davehorton:
I agree with @MrFabio .  @xquanluu please change it to `JAMBONES_REDIS_SENTINEL_SERVER_PASSWORD`

davehorton:
please update the realtime-helpers and rebase

PR #479, opened 2023-10-10
fix: transcribe 2 channels

xquanluu:
When dial transcribe is being use, feature server will request freeswitch 2 times for enable transcription for 2 different endpoints.

But due to this flag _speechHandlersSet, the seconds endpoint will not have speech credential for making connection to stt provider.

is there any idea why this flag is putting here and is there any regression if we remove this?

### DIFF
--- a/lib/tasks/transcribe.js
+++ b/lib/tasks/transcribe.js
@@ -175,8 +175,8 @@ class TaskTranscribe extends SttTask {
   }
 
   async _setSpeechHandlers(cs, ep, channel) {
-    if (ep.speechHandlersSet) return;
-    ep.speechHandlersSet = true;
+    if (ep._speechHandlersSet) return;
+    ep._speechHandlersSet = true;
     const opts = this.setChannelVarsForStt(this, this.sttCredentials, this.data.recognizer);
     switch (this.vendor) {
       case 'google':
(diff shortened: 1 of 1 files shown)


davehorton (commented):
It seems like the intent is to call this function only once per ep.  I think the bug was that it would only be called once, so the call to the second endpoint got ignored.  But your fix seems to allow it to be called unlimited times.  Shouldn't the fix be to ensure it is called only once per endpoint?  Also does `setChannelVarsForStt` need to be modified to take the ep as a param, so that it can act on either channel?

xquanluu:
fixed

lib/tasks/transcribe.js
```diff
@@ -175,8 +175,8 @@ class TaskTranscribe extends SttTask {
   }
 
   async _setSpeechHandlers(cs, ep, channel) {
-    if (this._speechHandlersSet) return;
-    this._speechHandlersSet = true;
+    if (ep.speechHandlersSet) return;
```
davehorton, line 178: please rename to `ep._speechHandlersSet`

davehorton (changes requested):

lib/tasks/transcribe.js
xquanluu: fixed

davehorton (approved):

Issue #472, opened 2023-09-27
Play and Say Verb in gather to have array of messages 


vdharashive:
gather to support play and say together as array and message to be played sequentially 

Also Say and play should support array with itself 

{
  "verb": "gather",
  "actionHook": "http://example.com/collect",
  "input": ["digits", "speech"],
  "bargein": true,
  "dtmfBargein": true,
  "finishOnKey": "#",
  "numDigits": 5,
  "timeout": 8,
  "recognizer": {
    "vendor": "Google",
    "language": "en-US",
    "hints": ["sales", "support"],
    "hintsBoost": 10
  },
  "sayplay": { 
  
  [
  say:{
    "text": "To speak to Sales press 1 or say Sales.  To speak to customer support press 2 or say Support",
    "synthesizer": {
      "vendor": "Google",
      "language": "en-US"
    },
  say:{
    "text": "To speak to Sales press 1 or say Sales.  To speak to customer support press 2 or say Support",
    "synthesizer": {
      "vendor": "Google",
      "language": "en-US"
    },
  play :{
   "url": "https://example.com/example.mp3",
   "actionHook": "/play/action"
  },
  
  say:{
    "text": "To speak to Sales press 1 or say Sales.  To speak to customer support press 2 or say Support",
    "synthesizer": {
      "vendor": "Google",
      "language": "en-US"
    },
 play :{
   "url": "https://example.com/example.mp3",
   "actionHook": "/play/action"
  }

  }
  
  ]
  }
}

PR #482, opened 2023-10-12
say verb support play audio in array

xquanluu:
https://github.com/jambonz/jambonz-feature-server/issues/472


lib/tasks/say.js
```diff
@@ -153,7 +162,13 @@ class TaskSay extends Task {
         }
       };
 
-      const arr = this.text.map((t) => generateAudio(t));
+      const arr = this.text.map((t) => {
```
davehorton, line 165: could we just make this
```
const arr = this.text.map((t) => this._validateURL(t) ? t : generateAudio(t);
```

davehorton (changes requested):

davehorton:
could you add a test case showing an array of mixed text and http urls to say

lib/tasks/say.js
davehorton: @xquanluu ?

xquanluu: fixed

xquanluu:
added testcase.

davehorton (approved):

PR #487, opened 2023-10-18
feat add http proxy to undicy

xquanluu:


davehorton (changes requested):
Could you add some test cases?  See the changes that I made to speech-utils github CI actions to spin up squid to be a local proxy

xquanluu:
Hi @davehorton when I testing this with our testsuite, it does not work at for Pool 
I have changed the http-request that when HTTP_PROXY is configured, each request will be send as single new proxy agent.

https://github.com/nodejs/undici/issues/1270

davehorton:
can you rebase?

lib/config.js
```diff
@@ -121,6 +121,9 @@ const HTTP_POOL = process.env.HTTP_POOL && parseInt(process.env.HTTP_POOL);
 const HTTP_POOLSIZE = parseInt(process.env.HTTP_POOLSIZE, 10) || 10;
 const HTTP_PIPELINING = parseInt(process.env.HTTP_PIPELINING, 10) || 1;
 const HTTP_TIMEOUT = 10000;
+const HTTP_PROXY_IP = process.env.JAMBONES_HTTP_PROXY_IP;
+const HTTP_PROXY_PORT = process.env.JAMBONES_HTTP_PROXY_PORT;
+const HTTP_PROXY_PROTOCOL = process.env.JAMBONES_HTTP_PROXY_PROTOCOL || 'http';
```
davehorton, line 126: I think the general case is we want to proxy both.  Do we need this variable?

xquanluu: this is to be used for traffic between feature server and squid, from squid to the internet (for example) should be http/https no matter

lib/utils/http-requestor.js
```diff
@@ -60,6 +73,18 @@ class HttpRequestor extends BaseRequestor {
       if (u.port) this.client = new Client(`${u.protocol}://${u.resource}:${u.port}`);
       else this.client = new Client(`${u.protocol}://${u.resource}`);
     }
+
+    if (NODE_ENV == 'test' && process.env.JAMBONES_HTTP_PROXY_IP) {
```
davehorton, line 77: why is this in test mode only?

xquanluu: In production: setGlobalDispatcher will be initiated only 1 time when feature server is boot.

When running in our testsuite, these HTTP PROXY env variables will be set on and off and we want to change global dispatcher whenever these env variables is changed.

davehorton (approved):

Issue #489, opened 2023-10-18
calls between authenticated users in the same sip realm should work even without an application


davehorton:
Say two users 'daveh' and 'hoan' have each successfully registered with the sip realm 'sip.jambonz.me'.  If either tries to dial the other, the call should be connected, even though neither has configured a default application to be used for sip devices.

Say daveh calls hoan.  The invite received by the feature server will include the information that this call is from a registered user, in the X-Authenticated-User header:
```
2023-10-18 09:07:48.227897 send 1751 bytes to udp/[10.0.35.194]:5070 at 09:07:48.227790:
INVITE sip:hoan@sip.jambonz.me SIP/2.0^M
Via: SIP/2.0/UDP 174.129.198.113;rport;branch=z9hG4bKc93jm05pX0vQH^M
Max-Forwards: 70^M
From: "500" <sip:daveh@174.129.198.113:5060>;tag=Ueej6yetNUFra^M
To: <sip:hoan@sip.jambonz.me>^M
Call-ID: a62d0a12-e838-123c-3c91-028d3aae2477^M
CSeq: 74256242 INVITE^M
Contact: <sip:10.0.35.194:5060>^M
User-Agent: JsSIP 3.10.0^M
Allow: INVITE, ACK, CANCEL, BYE, UPDATE, MESSAGE, OPTIONS, REFER, INFO, NOTIFY^M
Supported: timer, ice, replaces, outbound^M
Content-Type: application/sdp^M
Content-Length: 1040^M
X-Account-Sid: 9351f46a-678c-43f5-b8a6-d4eb58d131af^M
X-CID: 5641s4hsa27f5pdq5cpd^M
X-Forwarded-For: 107.115.5.130^M
X-Authenticated-User: daveh@sip.jambonz.me^M
X-Original-Number: hoan^M
```

The feature server should handle the invite by first doing what it currently does -- looking for an application for device calls.  If none is found, before returning this:
```
2023-10-18 09:07:48.233032 send 345 bytes to wss/[107.115.5.130]:31102 at 09:07:48.232952:
SIP/2.0 480 Temporarily Unavailable^M
Via: SIP/2.0/WSS pv0uuk3dp2vm.invalid;branch=z9hG4bK3479183;received=107.115.5.130;rport=31102^M
From: "500" <sip:daveh@sip.jambonz.me>;tag=fur5colm45^M
To: <sip:hoan@sip.jambonz.me>;tag=vQ7a8SZXj45ap^M
[...]

PR #491, opened 2023-10-20
Feat/devices call other

xquanluu:
https://github.com/jambonz/jambonz-feature-server/issues/489

### DIFF
--- a/lib/middleware.js
+++ b/lib/middleware.js
@@ -201,12 +201,8 @@ module.exports = function(srf, logger) {
               logger.debug(`There is no device app found, called client is registered,
  forwarding call to called client.`);
               app = {
-                // Dummy hook follow the feature server logic.
+                // Dummy hook to follow later feature server logic.
                 call_hook: {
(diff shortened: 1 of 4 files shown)


lib/middleware.js
```diff
@@ -191,8 +192,37 @@ module.exports = function(srf, logger) {
...
+            logger.debug({app}, `retrieved device calling app for realm ${sipRealm}`);
+          } else {
+            const calledAor = `${req.calledNumber}@${sipRealm}`;
+            const reg = await registrar.query(calledAor);
+            if (reg) {
+              logger.debug(`There is no device app found, called client is registered,
+ forwarding call to called client.`);
+              app = {
+                // Dummy hook to follow later feature server logic.
+                call_hook: {
```
davehorton, line 205: I guess something is required here?  It will never get used though right?

xquanluu: right, the rest of feature server logic need this info to initiate data, but never be used in p2p call.

davehorton (approved):

Issue #498, opened 2023-10-24
factor dynamic payloads into a separate file


davehorton:
We are starting to add a lot of code to middleware.js having to do with dynamic apps we are creating on the fly to support the webrtc client, like [here](https://github.com/jambonz/jambonz-feature-server/blob/3171b138f9abcb44c132633374189539887c8bf9/lib/middleware.js#L218), and we are likely to add more in the future and it is starting to clutter this file.  Let's factor those out into a new file under the 'data' folder that we can simply import.

PR #499, opened 2023-10-24
Fix/dynamic apps

xquanluu:
https://github.com/jambonz/jambonz-feature-server/issues/498

### DIFF
--- a/lib/apps-map.js
+++ b/lib/apps-map.js
@@ -34,7 +34,7 @@ const appsMap = {
 };
 
 const utilizeJambonzApp = (type, {account_sid, name, caller_id}) => {
-  const app = appsMap[type];
+  const app = {...appsMap[type]};
   app.account_sid = account_sid;
   switch (type) {
     case 'queue':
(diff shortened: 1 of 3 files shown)


lib/apps-map.js
```diff
@@ -0,0 +1,54 @@
...
+        {
+          type: 'user',
+          name: ''
+        }
+      ]
+    }]
+  }
+};
+
+const utilizeJambonzApp = (type, {account_sid, name, caller_id}) => {
```
davehorton, line 36: can you rename to createJambonzApp

lib/middleware.js
```diff
@@ -13,6 +13,7 @@ const listTaskNames = require('./utils/summarize-tasks');
 const {
   JAMBONES_MYSQL_REFRESH_TTL
 } = require('./config');
+const { utilizeJambonzApp } = require('./apps-map');
```
davehorton, line 16: can you change the file name from apps-map to dynamic-apps

davehorton (changes requested):

lib/apps-map.js
xquanluu: fixed

lib/middleware.js
xquanluu: fixed

davehorton (approved):

PR #514, opened 2023-11-01
feat disable direct p2p call by env variable

xquanluu:


lib/middleware.js
```diff
@@ -66,7 +66,7 @@ module.exports = function(srf, logger) {
       req.locals.queue_name = queue_name;
     }
     // check for call to registered user
-    if (req.locals.originatingUser) {
+    if (!JAMBONES_DISABLE_DIRECT_P2P_CALL && req.locals.originatingUser) {
```
davehorton, line 69: shouldnt this be in config.js with all env vars

davehorton (changes requested):

davehorton:
need to rebase

PR #515, opened 2023-11-02
feat assembly

xquanluu:


lib/tasks/transcribe.js
```diff
@@ -293,6 +294,20 @@ class TaskTranscribe extends SttTask {
...
+      case 'assemblyai':
+        this.bugname = 'assemblyai_transcribe';
+        ep.addCustomEventListener(AssemblyAiTranscriptionEvents.Transcription,
+          this._onTranscription.bind(this, cs, ep, channel));
+        ep.addCustomEventListener(AssemblyAiTranscriptionEvents.Connect,
+          this._onVendorConnect.bind(this, cs, ep));
+        ep.addCustomEventListener(AssemblyAiTranscriptionEvents.ConnectFailure,
+          this._onVendorConnectFailure.bind(this, cs, ep, channel));
+
+        /* if app sets deepgramOptions.utteranceEndMs they essentially want continuous asr */
```
davehorton, line 306: why is this here

lib/utils/transcription-utils.js
```diff
@@ -104,6 +105,9 @@ const stickyVars = {
   soniox: [
     'SONIOX_PROFANITY_FILTER',
     'SONIOX_MODEL'
+  ],
+  assemblyai: [
+    'ASSEMBLYAI_API_KEY'
```
davehorton, line 110: word boost should cleared as well I think

davehorton (changes requested):

lib/tasks/transcribe.js
xquanluu: copy paste mistake, removed them

lib/utils/transcription-utils.js
xquanluu: added now

davehorton (approved):

Issue #549, opened 2023-11-27
better support for opus codec when receiving calls from webrtc clients


davehorton:
Currently, calls from webrtc clients are connected using G711. It would be preferable to use OPUS codec for webrtc clients since this is a better-sounding audio codec.  Only one change is needed to use opus when it is favored by the client, and that is to change the codec preference setting in the sip profile in freeswitch from 'greedy' to 'generous'.  That change means that whatever the client prefers (from the SDP offer) we will try to accomodate.  Thus, when a webrtc client sends us an INVITE with opus preferred (ie first in the list of payload types in the SDP), we will choose that codec and use that.

All that is fine, and it works -- with the change to freeswitch calls from webrtc use opus and everything works fine while connected to the feature server.  However, when we then dial a B party we end up with no audio.  The problem is that on the B leg we send a list of codecs with G711 preferred, and the B leg then gets set up with G711.  We now have the A leg using OPUS and the B leg using G711.  So long as the media is anchored, it works because freeswitch will transcode.  However, when we release the media, there is no audio because we are not properly instructing rtpengine to transcode.

The desired state is that we set the codec negotiation setting to 'generous' and set up the A leg using their preferred codec.  Then, on an outdial to B, we should try if we can to match that same codec (e.g. OPUS on the B leg).  We want the INVITE going out from sbc-outbound to be offering whatever codec is being used on the A leg to be in the first position in the m= line of payload types.  If the remote B party then accepts that codec, that is great and we will have, in the case of a webrtc caller, OPUS being used all the way through.

However, the remote B party may choose to use a different codec.  If so, then we should have rtpengine on the B leg transcode that codec.  This will allow us to release the media anchor from freeswitch when possible, as we like to do.

In developing this feature, please proceed in stages:

1.  Change sip profile to generous and verify webrtc client (e.g. chrome extension) gets connected to "hello world" using OPUS.  Verify recording still works as before
2. Make webrtc-to-webrtc calling work where both legs are OPUS.  Verify audio is maintained after releasing media from freeswitch
3. Make webrtc-to-PSTN work where OPUS is on A leg but PSTN leg is using PMCU/PCMA.  Verify rtpengine is transcoding and that audio is maintained after releasing media from freeswitch

davehorton:
As a bonus, and while we are at it, we should also test with a client that prefers G.729 since from time to time this comes up as a request in bandwith-limited situations.

PR #550, opened 2023-11-28
fix release freeswitch media properly

xquanluu:
https://github.com/jambonz/jambonz-feature-server/issues/549

### DIFF
--- a/lib/tasks/dial.js
+++ b/lib/tasks/dial.js
@@ -790,9 +790,11 @@ class TaskDial extends Task {
     assert(cs.ep && sd.ep);
 
     try {
+      // Wait until we got new SDP from B leg to ofter to A Leg
       const aLegSdp = cs.ep.remote.sdp;
+      await sd.releaseMediaToSBC(aLegSdp, cs.ep.local.sdp);
       const bLegSdp = sd.dlg.remote.sdp;
-      await Promise.all[sd.releaseMediaToSBC(aLegSdp, cs.ep.local.sdp), cs.releaseMediaToSBC(bLegSdp)];
+      await cs.releaseMediaToSBC(bLegSdp);
       this.epOther = null;
       this.logger.info('Dial:_releaseMedia - successfully released media from freewitch');
     } catch (err) {


lib/utils/sdp-utils.js
```diff
@@ -12,6 +12,30 @@ const mergeSdpMedia = (sdp1, sdp2) => {
   return sdpTransform.write(parsedSdp1);
 };
 
+const isOpusFirst = (sdp) => {
+  const parsedSdp = sdpTransform.parse(sdp);
+  const idx = parsedSdp.media[0].rtp.findIndex((entry) => entry.codec === 'opus');
```
davehorton, line 17: since we reuse this function can we factor it out into a single line function, e.g
```
const getCodecPlacement = (sdp, codec ) => sdpTransform.parse(sdp).media[0].rtp.findIndex((e) => e.codec === codec):
```

lib/utils/sdp-utils.js
```diff
@@ -12,6 +12,30 @@ const mergeSdpMedia = (sdp1, sdp2) => {
...
+  return idx === 0;
+};
+
+const opusFirst = (sdp) => {
+  const parsedSdp = sdpTransform.parse(sdp);
+  // Find the index of the OPUS codec
+  const opusIndex = parsedSdp.media[0].rtp.findIndex((entry) => entry.codec === 'opus');
+
+  // Move OPUS codec to the beginning
+  if (opusIndex !== -1) {
```
davehorton, line 27: why are we comparing to -1 (not found)?  What if opusIndex is 0?  Then we should just leave it alone.  What if opusIndex is 3?  Then we should be moving it, but it seems we would be leaving it there

lib/utils/sdp-utils.js
```diff
@@ -12,6 +12,30 @@ const mergeSdpMedia = (sdp1, sdp2) => {
   return sdpTransform.write(parsedSdp1);
 };
 
+const isOpusFirst = (sdp) => {
+  const parsedSdp = sdpTransform.parse(sdp);
+  const idx = parsedSdp.media[0].rtp.findIndex((entry) => entry.codec === 'opus');
+  return idx === 0;
+};
+
+const opusFirst = (sdp) => {
```
davehorton, line 21: please rename this function to 'makeOpusFirst'

davehorton (changes requested):

lib/utils/sdp-utils.js
xquanluu: right, we just need to move if the index of opus > 0. 
Fixed

lib/utils/sdp-utils.js
xquanluu: fixed

lib/utils/sdp-utils.js
```diff
@@ -12,19 +12,19 @@ const mergeSdpMedia = (sdp1, sdp2) => {
...
-  const idx = parsedSdp.media[0].rtp.findIndex((entry) => entry.codec === 'opus');
-  return idx === 0;
+  return getCodecPlacement(sdp, 'opus') === 0;
 };
 
-const opusFirst = (sdp) => {
+const makeOpusFirst = (sdp) => {
   const parsedSdp = sdpTransform.parse(sdp);
   // Find the index of the OPUS codec
   const opusIndex = parsedSdp.media[0].rtp.findIndex((entry) => entry.codec === 'opus');
```
davehorton, line 24: I dont understand why we arent calling `getCodecPlacement` here?

lib/utils/sdp-utils.js
davehorton: see my comment below, I don't understand your point

lib/utils/sdp-utils.js
xquanluu: because we use parsedSdp result, calling getCodecPlacement here will make the code parse SDP 2 times for nothing

davehorton: I see.  Ok then perhaps change the function to accept a parsed SDP result instead of the raw SDP

lib/utils/sdp-utils.js
```diff
@@ -12,16 +12,21 @@ const mergeSdpMedia = (sdp1, sdp2) => {
   return sdpTransform.write(parsedSdp1);
 };
 
-const getCodecPlacement = (sdp, codec) => sdpTransform.parse(sdp).media[0].rtp.findIndex((e) => e.codec === codec);
+const getCodecPlacement = (parsedSdp, codec) => {
+  if (!!parsedSdp && !!parsedSdp.media && parsedSdp.media.length > 0) {
+    return parsedSdp.media[0].rtp.findIndex((e) => e.codec === codec);
```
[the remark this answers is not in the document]
xquanluu: fixed

davehorton (approved):

Issue #541, opened 2023-11-17
Feature Request: elevenlabs options


Catharsis68:
We'd have the requirement to be able to configure the TTS provider `elevenelabs` similar to other providers via an `elevenlabsOptions` object.

According to [elevenlabs docs](https://elevenlabs.io/docs/api-reference/text-to-speech-stream) we'd like the following properties:

- model_id
- voice_id
- optimize_streaming_latency
- voice_settings
  - similarity_boost
  - stability
  - style
  - use_speaker_boost

davehorton:
@xquanluu is this still open?

davehorton:
this also requires changes to webapp as well

davehorton:
ok lets work this.  They specifically have a new model that is supposed to be faster - turbo v2.  We should support it and make it the default in the webapp

xquanluu:
The feature is deployed to jambonz.me

PR #553, opened 2023-11-30
support elevenlabs options

xquanluu:
https://github.com/jambonz/jambonz-feature-server/issues/541

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,7 +15,7 @@
         "@jambonz/http-health-check": "^0.0.1",
         "@jambonz/mw-registrar": "^0.2.4",
         "@jambonz/realtimedb-helpers": "^0.8.7",
-        "@jambonz/speech-utils": "^0.0.26",
+        "@jambonz/speech-utils": "^0.0.29",
         "@jambonz/stats-collector": "^0.1.9",
         "@jambonz/time-series": "^0.2.8",
         "@jambonz/verb-specifications": "^0.0.46",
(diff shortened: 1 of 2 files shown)


Catharsis68:
We'd have the need to pass these options via `Config verb` - similar to e.g `deepgramOptions` but within the `synthesizer` property.  Or we could use the existing  `[synthesizer.options](https://github.com/jambonz/verb-specifications/blob/f860074a244373efeabc177e92273ce874b71867/specs.json#L494C12-L494C13)` property

xquanluu:
Thanks @Catharsis68 for the review, I have updated the code to get the options from synthezier.options.

this is deployed in jambonz.me

paulotelles:
@xquanluu 

The main goal here is to have the option to change those configs during the call like
`

{
      "config":{
        "synthesizer":{
          "vendor":"microsoft",
          "language":"en-US",
          "voice":"en-US-AriaNeural"
         "options": {
             "stability": 0.5,
             "similarity_boost": 0.5
             "model": turbo_v2
             "style": 1
             "use_speaker_boost": "true"
        }
}
`

xquanluu:
Hi @paulotelles the options is not being used for only elevenlabs, and for me it should be generic enough that later if there is something change at elevenlabs we don't need to come here and fix the format of options again. the application just very simple update the options to follow the new rule

Currently, You have to provide elevenlabs configuration to options. please rethink.

davehorton (approved):

PR #572, opened 2023-12-19
fix enqueue waithook on ws hold the session

xquanluu:


davehorton:
can you describe why this fix is needed and how it works?

xquanluu:
For enqueue waitHook, it create new instance of ConfirmCallSession to execute play/say/pause.
However ConfirmCallSession is extended from CallSession class that if the callHook is a ws connection, after finish all tasks ConfirmCallSession will be held to wait for another command from ws that make ConfirmCallSession cannot be finished and enqueue verb can move forward to execute next round of waitHook.

davehorton (approved):

PR #575, opened 2023-12-22
fix config and stt task for altLanguages

xquanluu:


davehorton (approved):

davehorton:
was this tested successfully @xquanluu ?

Issue #568, opened 2023-12-15
Add support for "background transcribe"


davehorton:
We currently have the notion in the code of "background listen" and "background gather".  In a similar fashion, I would like to extend the `config` verb to start or stop a background transcription task.  There are many situations where it may be desirable to transcribe a call while doing something else, but currently transcribe is a blocking verb.

So we should add a `transcribe` property to the `config` verb, where `transcribe` has properties to start or stop a transcription.  It should be possible to transcribe only the caller or both channels.  Note that it shall be possible to be doing a background transcription, and then later in the same call a `gather` or even a blocking`transcribe` that does not impact the background transcription.  It shall be possible to do only one background transcription on a session at a time though.

PR #576, opened 2023-12-22
Transcribe background task

xquanluu:
https://github.com/jambonz/jambonz-feature-server/issues/568


lib/tasks/gather.js
```diff
@@ -300,27 +301,27 @@ class TaskGather extends SttTask {
     const opts = this.setChannelVarsForStt(this, this.sttCredentials, this.data.recognizer);
     switch (this.vendor) {
       case 'google':
-        this.bugname = 'google_transcribe';
+        this.bugname = `${this.bugname_prefix}google_transcribe`;
```
davehorton, line 304: are we guaranteed to always have `bugname_prefix` property of this?  If not we would we have "undefined_google_transcribe"?  If so I would rather have `${this.bugname || ''}google_transcribe`

davehorton (commented):
can we change the name to background-task-manager.js and move it into the utils folder

lib/tasks/gather.js
xquanluu: By default is empty string:
https://github.com/jambonz/jambonz-feature-server/pull/576/files#diff-540439258e43bbda786b9a2d537d8b5c87ec55fc1f92c69c291861368f5fa94dR53

davehorton (approved):

PR #581, opened 2023-12-28
wip, create outbound dial from webhook ws

xquanluu:


lib/session/call-session.js
```diff
@@ -969,6 +970,91 @@ class CallSession extends Emitter {
...
+      await this._validateCreateCall(opts);
+      const resp = bent('POST', 201)(restDialUrl, opts);
+      this.logger.info(resp.body, 'successfully create outbound call');
+      return resp.body;
+    } catch (err) {
+      this.logger.error(err, 'failed to create outbound call from ' + this.callSid);
+    }
+  }
+
+  async _validateCreateCall(opts) {
```
davehorton, line 989: why do we need to validate here?  The endpoint we are calling `/createCall` will do this validation.  So it seems duplicated...can't we just capture the error response (typically a 400 Bad Request) and send it back over the websocket in some form?

davehorton (changes requested):

vdharashive:
@xquanluu is this completed ?

davehorton:
@xquanluu @vdharashive is this PR tested and ready for merge?

xquanluu:
From my site, It's tested and working together with conference example app.

xquanluu:
But wait me to clean up the code as you mentioned above

xquanluu:
Hi @davehorton I have clean up the code and retested on jambonz.me

Please help me to review the code one again

lib/session/call-session.js
xquanluu: I removed the part of code for data validation, I just now keep the code for transforming data that feature server can understand and process the outbound call

davehorton (approved):

PR #583, opened 2023-12-28
fix hangup headers

xquanluu:


lib/session/call-session.js
```diff
@@ -1883,7 +1883,7 @@ Duration=${duration} `
   wrapDialog(dlg) {
     dlg.connectTime = moment();
     const origDestroy = dlg.destroy.bind(dlg);
-    dlg.destroy = () => {
+    dlg.destroy = (opts, callback) => {
```
davehorton, line 1886: I dont think we should include `callback`.  Its optional, if the caller wants a callback rather than a Promise.  But we exclusively use Promises

lib/session/call-session.js
```diff
@@ -1892,7 +1892,7 @@ Duration=${duration} `
         this.emit('callStatusChange', {callStatus: CallStatus.Completed, duration});
         this.logger.debug('CallSession: call terminated by jambonz');
         this.rootSpan.setAttributes({'call.termination': 'hangup by jambonz'});
-        origDestroy().catch((err) => this.logger.info({err}, 'CallSession - error destroying dialog'));
+        origDestroy(opts, callback).catch((err) => this.logger.info({err}, 'CallSession - error destroying dialog'));
```
davehorton, line 1895: ditto above

davehorton (changes requested):

davehorton:
also fix failing tests

davehorton (approved):

Issue #573, opened 2023-12-19
Bug: after pause transcript , trancripts are still generating


vdharashive:
Blow log shows there is a restart on the transcript due to no audio , even after pause transcript start flowing after restart 

way to reproduce is before keep call on mute and then issue pause , transcript will start flowing

DEBUG [07:41:35.464] (402046): received command
    callId: "a428d3bc-181b-123d-05a6-0eb94b3c8831"
    callSid: "3d8686d3-f57e-4ec9-b94e-ca73a31d4ca3"
    accountSid: "c8e2ffe2-e6d1-46a7-9a63-1a662bc0dfc2"
    callingNumber: "+18448735673"
    calledNumber: "+12055339731"
    traceId: "c9374f9f0b64235d992edeb61bd17513"
    command: "transcribe:status"
    call_sid: "30a66491-ba17-4aec-9db6-ce010a75019e"
    queueCommand: true
    data: {
      "transcribe_status": "pause"
    }
INFO [07:41:35.464] (402046): CallSession:_onCommand - received command
    callId: "a428d3bc-181b-123d-05a6-0eb94b3c8831"
    callSid: "3d8686d3-f57e-4ec9-b94e-ca73a31d4ca3"
    accountSid: "c8e2ffe2-e6d1-46a7-9a63-1a662bc0dfc2"
    callingNumber: "+18448735673"
    calledNumber: "+12055339731"
    traceId: "c9374f9f0b64235d992edeb61bd17513"
    command: "transcribe:status"
    queueCommand: true
    data: {
      "transcribe_status": "pause"
    }
INFO [07:41:35.465] (402046): TaskTranscribe:updateTranscribe status pause
    callId: "a428d3bc-181b-123d-05a6-0eb94b3c8831"
    callSid: "3d8686d3-f57e-4ec9-b94e-ca73a31d4ca3"
    accountSid: "c8e2ffe2-e6d1-46a7-9a63-1a662bc0dfc2"
    callingNumber: "+18448735673"
    calledNumber: "+12055339731"
    traceId: "c9374f9f0b64235d992edeb61bd17513"
DEBUG [07:41:36.530] (402046): TaskTranscribe:_onNoAudio restarting transcription on channel 1
    callId: "a428d3bc-181b-123d-05a6-0eb94b3c8831"
    callSid: "3d8686d3-f57e-4ec9-b94e-ca73a31d4ca3"
[...]

PR #584, opened 2023-12-28
#573 address race condition in pause/resume recording

davehorton:


vdharashive:
@davehorton can this be rebased with latest commit , i have noticed this issue

PR #586, opened 2023-12-29
fix pause resume background transcribe

xquanluu:


xquanluu:
Tested on jambonz.me

lib/session/call-session.js
```diff
@@ -1059,6 +1059,11 @@ class CallSession extends Emitter {
    * @param {string} opts.transcribe_status - 'pause' or 'resume'
   */
   async _lccTranscribeStatus(opts) {
+    if (this.backgroundTaskManager.isTaskRunning('transcribe')) {
+      this.backgroundTaskManager.getTask('transcribe').updateTranscribe(opts.transcribe_status);
+    } else {
+      this.logger.info('CallSession:_lccTranscribeStatus - there is no background transcribe task');
```
davehorton, line 1065: this will be by far the common case, and no need to log any thing

davehorton (changes requested):

davehorton:
@xquanluu should we merge this?

xquanluu:
Yes, I tested this.

The question for you is should we allow LCC transcribe_status be able to stop both background transcribe and foreground transcribe together.

davehorton:
> Yes, I tested this.
> 
> The question for you is should we allow LCC transcribe_status be able to stop both background transcribe and foreground transcribe together.

So right now it only pauses the background transcribe, right?

xquanluu:
> > Yes, I tested this.
> > The question for you is should we allow LCC transcribe_status be able to stop both background transcribe and foreground transcribe together.
> 
> So right now it only pauses the background transcribe, right?

Right now, it's only pauses the foreground transcribe

davehorton:
ah, ok I see.  Well, I think we should wait on this PR until we can make it pause both/all transcribes.  I think the common use case is that some PII data is being spoken and we wouldn't want it to appear on any transcripts

xquanluu:
I see, If that's the case then the PR is ready for merging, this PR will pause and resume both background and foreground transcribe

PR #588, opened 2024-01-04
fix stt default vendor cannot be mapped to correct value

xquanluu:


vdharashive:
its working now as expected

davehorton (approved):

Issue #589, opened 2024-01-05
Feat: apply config synthesizer.options in say task


Catharsis68:
## Feature Request

**Use Case:** 

We want to apply advanced elevenlabs synthesizer options on a "global" level for a whole call via config verb.

**Actual:**
It is possible to apply the elevenlabs options via say task. `synthesizer.options` are being used when calling `synthAudio`
    - with this approach, we'd need to define the options on every say task in our flow editor where we generate the say task payload

**Goal:**
We can send the same `synthesizer.options` at the beginning of a call via `config` task and if defined, they will be used for `synthAudio`. In the case the say task has it's own `synthesizer.options`, this should overwrite the "config" settings.

xquanluu:
Hi @Catharsis68 I have tested this PR on jambonz.me, 

Please help to test on your system please

xquanluu:
I close this ticket as I remember @Catharsis68  already provided the PR for this

PR #590, opened 2024-01-05
feat override synthesizer options from config

xquanluu:
https://github.com/jambonz/jambonz-feature-server/issues/589


xquanluu:
Tested on jambonz.me

lib/tasks/say.js
```diff
@@ -68,6 +68,12 @@ class TaskSay extends Task {
     const salt = cs.callSid;
 
     let credentials = cs.getSpeechCredentials(vendor, 'tts', label);
+
+    // override Synthesizer options from config verb
+    if (cs.synthesizer && cs.synthesizer.options) {
```
Catharsis68, line 73: @xquanluu  I would add these conditions: 
- call session synthesizer options are only applied, when there is no say.synthesizer.options
- at least one "key" must be present in call session synthesizer options 
```
if (Object.keys(this.options).length === 0 && cs.synthesizer && cs.synthesizer.options && Object.keys(cs.synthesizer.options).length > 0) {
```

xquanluu: Thanks @Catharsis68 for the review:
1/ call session synthesizer options are only applied, when there is no say.synthesizer.options
--> This seem to me we violate the rule for config verb to override configuration from other verb.

2/ at least one "key" must be present in call session synthesizer options
--> Do you know any case that user use config verb to put {} empty configuration, IMHO I think there is 2 normal case there is  synthesizer.options and there is no synthesizer.options?

davehorton: I agree with @Catharsis68 on this one -- the config provides session-level settings, but if a verb specifies something different, that should take preference for that verb.  Given that, I think we can merge this @xquanluu ?

lib/tasks/say.js
```diff
@@ -70,7 +70,10 @@ class TaskSay extends Task {
     let credentials = cs.getSpeechCredentials(vendor, 'tts', label);
 
     // override Synthesizer options from config verb
-    if (cs.synthesizer && cs.synthesizer.options) {
+    if (!Object.keys(this.options).length != 0 &&
```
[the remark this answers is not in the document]
davehorton: if not length does not equal zero.....can't that be made into a simple test: if length equals zero?

davehorton (changes requested):

PR #593, opened 2024-01-06
Fix/dual legs transcribe race condition

xquanluu:


lib/tasks/transcribe.js
```diff
@@ -263,7 +264,8 @@ class TaskTranscribe extends SttTask {
           throw new Error(`Invalid vendor ${this.vendor}`);
         }
     }
-
+    // save dedicated bugname for each endpoint
+    ep.transcribe_bugname = `${bugname}_${Date.now()}`;
```
davehorton, line 268: can we use a simple integer counter rather than Date.now() ?  It might be easier to debug if necessary

davehorton (changes requested):

lib/tasks/transcribe.js
xquanluu: fixed

xquanluu: No We can't, due to issue that new transcribe verb can have the same name again to the old transcribe verb is being killed

lib/tasks/transcribe.js
```diff
@@ -265,7 +266,8 @@ class TaskTranscribe extends SttTask {
         }
     }
     // save dedicated bugname for each endpoint
-    ep.transcribe_bugname = `${bugname}_${Date.now()}`;
+    ep.transcribe_bugname = `${bugname}_${channel}_${this.bugname_counter}`;
```
davehorton: maybe just 
```
ep.transcribe_bugname = `${bugname}_${channel}_${this.bugname_counter++}`;
```

davehorton (changes requested):

lib/tasks/transcribe.js
davehorton: I'm confused though...seems like did make this change below?

lib/tasks/transcribe.js
xquanluu: No, we can't use this, 

first verb will have transcribe_bugname_1_0
LCC to move the call to new transcribe verb in listen will have same bug name again transcribe_bugname_1_0

in the meanwhile the first transcribe is still being closed, the 2nd transcribe save private data to same bug name again

Still Date.now() is a must to make sure every single time we have different bugname

davehorton (approved):

PR #594, opened 2024-01-08
Fix/gather timeout

xquanluu:
Fix gather cannot be timeout if missing `minDigits`

https://github.com/jambonz/jambonz-feature-server/issues/597


davehorton:
can you open an issue and describe this bug?

xquanluu:
Opened issue

lib/tasks/gather.js
```diff
@@ -35,6 +35,7 @@ class TaskGather extends SttTask {
 
     /* timeout of zero means no timeout */
     this.timeout = this.timeout === 0 ? 0 : (this.timeout || 15) * 1000;
+    this.interDigitTimeout = (this.interDigitTimeout > 0 ? this.interDigitTimeout : 0) * 1000;
```
davehorton, line 38: was this changed needed to fix this bug?  Or is it an unrelated fix?

xquanluu: this is to make sure there is value for interDigitTimeout that later we can do compare this. interDigitTimeout <= 0.

But I think this is not needed.

I added new changes and retested on jambonz.me

davehorton: Yes I have a feeling we are over-complicating things here.  I will have more time to review later today

davehorton (approved):

lib/tasks/gather.js
davehorton: where are we with this PR @xquanluu ?

xquanluu:
I remembered that you gave me some suggestion to solve this problem, I applied and retested on jambonz.me, and it was working at that time.

It's good to be merged

davehorton (approved):

PR #600, opened 2024-01-09
Revert "Fix/dual legs transcribe race condition (#593)"

davehorton:
This reverts commit 9d70ed96a11d7e695fa8d22d8f5da9440072c201.  That commit was faulty and caused a regression bug


vdharashive:
@davehorton @xquanluu i had notice that after applying this https://github.com/jambonz/jambonz-feature-server/commit/9d70ed96a11d7e695fa8d22d8f5da9440072c201. commit has fixied issues for me. other wise there was only transcript was coming for single channel .

PR #602, opened 2024-01-10
add user-agent to http and ws requestor

xquanluu:


lib/utils/ws-requestor.js
```diff
@@ -233,6 +229,9 @@ class WsRequestor extends BaseRequestor {
         maxRedirects: 2,
         handshakeTimeout,
         maxPayload: JAMBONES_WS_MAX_PAYLOAD ? parseInt(JAMBONES_WS_MAX_PAYLOAD) : 24 * 1024,
+        headers: {
```
davehorton, line 232: I would rrather no user agent header if env is not set

davehorton (changes requested):

davehorton (approved):

PR #612, opened 2024-01-15
feat: allow update azure endpoint ID from recognizer property

xquanluu:
https://jambonz.freshdesk.com/support/tickets/24


davehorton (commented):
this change can not be accepted until a PR for verb-specs has been accepted, correct?

xquanluu:
Hi @davehorton @vdharashive 

For tts, we can use sythesizer.options.deploymentId to override MS endpoint ID
For STT, we can use recognizer.azureSttEndpointId

Those parameter is already supported by verb specification. So there is no need to have another PR.

davehorton (approved):

PR #615, opened 2024-01-16
Feature/add synth options call session

Catharsis68:
Based on this PR #590 

- Add `speechSynthesisOptions` from config verb to a Call Session
- In say task: if defined, apply the `speechSynthesisOptions` to `this.options` - but let them get overwritten by `say.synthesizer.options`


davehorton:
@xquanluu do you have an opinion on this PR?

xquanluu:
For me I don't know the issue related to this, but from the code, he wants to override say verb synthesizer options from config verb if say verb syntheizer.options is missing

Issue #617, opened 2024-01-17
call is released for await ws next task if there is ws command without new tasks initiated


xquanluu:
1/ The call was in waiting next task as ws is used and there is no more task to execture
2/ The call was received sip:requirest command that make jambonz release call await for next task
3/ call session for the call was released due to await  any more.

PR #619, opened 2024-01-18
fixed call hangup as call is await for new task and received ws command

xquanluu:
#617


vdharashive:
@davehorton Please merge this its tested

davehorton (approved):

PR #629, opened 2024-01-25
Tts/elevenlabs streaming

xquanluu:


lib/session/call-session.js
```diff
@@ -2033,8 +2034,12 @@ Duration=${duration} `
   }
 
   _configMsEndpoint() {
```
xquanluu, line 2036: https://github.com/jambonz/jambonz-feature-server/blob/main/lib/utils/place-outdial.js#L326

We might want to update this.

davehorton: we basically have two duplicate functions?  Is there no way to centralize that code?

PR #630, opened 2024-01-29
Fix dequeue

avoylenko:
The **retrieveByPatternSortedSet** returns an array that is always true even though its empty. As result the url is an empty string that breaks the dequeue task.

How to reproduce:
- Run the dequeue tasks for unexisting call


davehorton:
@xquanluu can you review

xquanluu:
Thanks @avoylenko  for the PR

Hi @davehorton I'm good with the PR, I have tested this on jambonz.me and it's working as expected.

davehorton (approved):

PR #632, opened 2024-01-31
fix default gather input is digits and gather dtmf should not require speech

xquanluu:
#631


davehorton (approved):

xquanluu:
Tested with transcribe and gather (digits or speech), all are working as expected

PR #633, opened 2024-01-31
fix ws reconnect does not send verb:hook data

xquanluu:


davehorton (approved):

xquanluu:
It's ready, I tested on jambonz.me, waiting for the test from customer

PR #634, opened 2024-01-31
Only Start Intedigit timeout if input is digits

xquanluu:


davehorton (commented):
I dont see any changes here other than insertion of leading spaces?  What am I missing

xquanluu:
I added the line 293 on that file

lib/tasks/gather.js
```diff
@@ -287,7 +287,10 @@ class TaskGather extends SttTask {
       this._startFinalAsrTimer();
       return;
     }
-    if (!resolved && this.interDigitTimeout > 0 && this.digitBuffer.length >= this.minDigits) {
+    if (!resolved &&
+        this.interDigitTimeout > 0 &&
+        this.digitBuffer.length >= this.minDigits &&
```
davehorton, line 292: I don't think the addition of the check of this.input should be necessary.  There should never be any digits in the buffer if we are not collecting digits.

davehorton: so how did digits get in the buffer?  That is what we should fix

davehorton: @xquanluu any thoughts on this?

Issue #637, opened 2024-02-06
dialMusic should loop


sammachin:
using the `dialMusic` param of the `dial` verb to play ringback tone while a call is connecting, it  only play the mp3 once, 

I would have expected it to loop,

This means I need to create a 60sec MP3 instead of a 3sec one,

PR #640, opened 2024-02-11
Loop dialMusic

avoylenko:
Play the dial music until the task is cancelled or connected. Hopefully solves https://github.com/jambonz/jambonz-feature-server/issues/637


davehorton:
@avoylenko have you tested this?

lib/tasks/dial.js
```diff
@@ -196,7 +196,11 @@ class TaskDial extends Task {
       else {
         this.epOther = cs.ep;
         if (this.dialMusic && this.epOther && this.epOther.connected) {
-          this.epOther.play(this.dialMusic).catch((err) => {});
+          (async() => {
```
davehorton, line 199: what happens if an error is thrown from the play command?  Will it be an uncaught exception and crash the program?  This would need to be tested

avoylenko: The exception is being handled by the catch section attached to play method. The main issue could be if the play method will be throwing the exception in a loop and we will end with endless loop blocking all other tasks.

avoylenko: Not sure that this is a realistic scenario

avoylenko: Maybe we could wait inside the catch block to workaround this

avoylenko: @davehorton I've added a short delay inside the catch section, just in case we receive a malformed url for dialMusic.

avoylenko:
@davehorton is there something blocks you from merging this?

davehorton:
can you rebase and I will give another review

PR #670, opened 2024-02-29
allow custom header on pause, resume recording

xquanluu:
waiting them to test on real env

### DIFF
--- a/lib/session/call-session.js
+++ b/lib/session/call-session.js
@@ -622,8 +622,11 @@ class CallSession extends Emitter {
       const res = await this.dlg.request({
         method: 'INFO',
         headers: {
-          'X-Reason': 'pauseCallRecording'
-        }
+          'X-Reason': 'pauseCallRecording',
+          ...(this.recordOptions.headers && {'Content-Type': 'application/json'})
+        },
+        // Siprect Client is initiated from startCallRecording, so just need to pass custom headers in startRecording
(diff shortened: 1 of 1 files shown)


xquanluu:
They confirmed feature is working for re-invite custom headers on recording command

lib/session/call-session.js
```diff
@@ -622,8 +622,11 @@ class CallSession extends Emitter {
       const res = await this.dlg.request({
         method: 'INFO',
         headers: {
-          'X-Reason': 'pauseCallRecording'
-        }
+          'X-Reason': 'pauseCallRecording',
+          ...(this.recordOptions.headers && {'Content-Type': 'application/json'})
```
davehorton, line 626: are we guaranteed that this.recordOptions will exist?

lib/session/call-session.js
```diff
@@ -622,8 +622,11 @@ class CallSession extends Emitter {
...
+        },
+        // Siprect Client is initiated from startCallRecording, so just need to pass custom headers in startRecording
```
davehorton, line 628: minor typo "Siprect"

lib/session/call-session.js
xquanluu: https://github.com/jambonz/jambonz-feature-server/blob/main/lib/session/call-session.js#L549

before executing recording command, recordOptions = data from recording command. so it always available for recording command

davehorton (approved):

PR #679, opened 2024-03-11
fix actionHookDelay feature is not working properly if there is no de…

xquanluu:
…fined actions


vdharashive:
@xquanluu In this case when a call is terminated its comes as caller hangup rather it should be jambonz

xquanluu:
Hi @vdharashive  just pushed new code for terminated by jambonz/caller fix. Tested on jambonz.me, please give it a try.

vdharashive:
@xquanluu @davehorton this can be merged its working as expected

davehorton (approved):

PR #684, opened 2024-03-14
gather verb should clean dtmf listerner even dtmfBargein=false

xquanluu:


lib/tasks/gather.js
```diff
@@ -965,10 +965,8 @@ class TaskGather extends SttTask {
     // If bargin is false and ws application return ack to verb:hook
     // the gather should not play any audio
     this._killAudio(this.cs);
```
davehorton, line 967: this listener is removed in the `kill` method which is always called so I dont see the value of doing this here

davehorton (changes requested):

PR #686, opened 2024-03-17
gather verb should clean dtmf listerner even dtmfBargein=false

xquanluu:


xquanluu:
Hi @davehorton  finally I can reproduce the issue for PossibleEventEmittermemoryleakdetected.11dtmflistenersaddedto

Steps:

1/ Use gather with input digits
2/ Provide dtmf and make sure gather resolve with the dtmf input
3/ Gather send verb hook to WS application which then return ack, no verbs
4/ after 4 seconds ws application send redirect command for new gather
5/ loop back to step 1 and do that loop for 11 times.

The issue is:

Kill() function inside of each task is called when call-session release the call or replace current task list by another task list. Incase the task completed by itself, kill function will never be called, that make dtmf listener will never be called.

This PR is about to clean up dtmf event if resolve function is called.

davehorton:
@xquanluu I am unclear on the status of this PR.  Did we decide it was not needed, or still needed?

xquanluu:
Hi @davehorton  it is needed to be merged

davehorton (approved):

PR #695, opened 2024-03-27
support listen verb support bidirectionalAudioSampleRate

xquanluu:
https://github.com/jambonz/freeswitch-modules/issues/25

### DIFF
--- a/lib/tasks/listen.js
+++ b/lib/tasks/listen.js
@@ -13,7 +13,7 @@ class TaskListen extends Task {
 
     [
       'action', 'auth', 'method', 'url', 'finishOnKey', 'maxLength', 'metadata', 'mixType', 'passDtmf', 'playBeep',
-      'sampleRate', 'timeout', 'transcribe', 'wsAuth', 'disableBidirectionalAudio'
+      'sampleRate', 'timeout', 'transcribe', 'wsAuth', 'disableBidirectionalAudio', 'bidirectionalAudioSampleRate'
     ].forEach((k) => this[k] = this.data[k]);
 
     this.mixType = this.mixType || 'mono';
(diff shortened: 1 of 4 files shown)


lib/tasks/listen.js
```diff
@@ -133,7 +133,8 @@ class TaskListen extends Task {
       mixType: this.mixType,
       sampling: this.sampleRate,
       ...(this._bugname && {bugname: this._bugname}),
-      metadata
+      metadata,
+      bidirectionalAudio: this.bidirectionalAudio || {}
```
davehorton, line 137: disableBidirectionalAudio is there for legacy purposes, so we have to keep it.  But I think in this listen.js code we should immediately set `bidirectionalAudio.enabled` here and then in the rest of the code refer only to that, ie.right after the line above do something like
```
if ('enabled' not in this.bidirectionalAudio) {
   this.bidirectionalAudio.enabled = this. disableBidirectionalAudio === true ? false : true;
}
```

davehorton (changes requested):

davehorton:
listen tests are failing

lib/tasks/listen.js
davehorton: are you going to address this?

xquanluu: Yes doing it now

xquanluu: fixed

davehorton:
I think we need to explicitly handle the case where the feature has been updated to have the bidirectional streaming feature but the freeswitch has not.  Could you make sure this results in the feature-server logging an error message and ending the listen task, proceeding to the next task

xquanluu:
> I think we need to explicitly handle the case where the feature has been updated to have the bidirectional streaming feature but the freeswitch has not. Could you make sure this results in the feature-server logging an error message and ending the listen task, proceeding to the next task

The issue is, the forkAudioStart will not be failed on feature server due to freeswitch still be able to open ws to destination as normal, but bugname or metadata could be wrong due to we added bidirectionalAudio params.

davehorton:
can you rebase to main?

xquanluu:
rebased

davehorton:
I updated the freeswitch image, test passing now, can you rebase again @xquanluu ?

davehorton (approved):

PR #712, opened 2024-04-09
kill play task if bot responds verbs while actionHook delay is enabled

xquanluu:


vdharashive:
@xquanluu and @davehorton this is working as expected can this me merged in 0.8.5

PR #733, opened 2024-04-26
support deepgram use single connection

xquanluu:
https://github.com/jambonz/freeswitch-modules/pull/56


davehorton:
recall that I wanted no changes to feature server for this feature

PR #736, opened 2024-05-01
feat tts stream fallback

xquanluu:


lib/tasks/say.js
```diff
@@ -239,17 +242,14 @@ class TaskSay extends TtsTask {
...
-    } catch (error) {
+    const startFallback = async(error) => {
       if (fallbackVendor && this.isHandledByPrimaryProvider && !cs.hasFallbackTts) {
         this.notifyError(
           { msg: 'TTS error', details:`TTS vendor ${vendor} error: ${error}`, failover: 'in progress'});
         this.isHandledByPrimaryProvider = false;
         cs.hasFallbackTts = true;
         this.logger.info(`Synthesize error, fallback to ${fallbackVendor}`);
-        filepath = await this._synthesizeWithSpecificVendor(cs, ep,
+        this._filepath = await this._synthesizeWithSpecificVendor(cs, ep,
```
davehorton, line 252: why is it necessary to make this an instance variable now?

xquanluu: You are right, I was bad here, revert it to local variable

PR #742, opened 2024-05-03
fix dead lock in say verb while waiting playback-stop and say verb is…

xquanluu:
If say verb is waiting for playstop-event happens, and then new command that cause replaceApplication, the ongoing say verb will be killed, but the say.exec cannot be released due to it's still wait for playstop-callback.


vdharashive:
@xquanluu fix is working thanks

davehorton (approved):

vdharashive:
@davehorton can you please merge this

PR #769, opened 2024-05-31
Loop dial music until connected

avoylenko:
Play the dial music until the task is cancelled or connected. Hopefully solves https://github.com/jambonz/jambonz-feature-server/issues/637


avoylenko:
@davehorton is there something that blocks you from approving?

davehorton:
just super busy with customers.  Will have a look at this when I can this week.

PR #770, opened 2024-06-01
Fix/verbio stt

xquanluu:
Fix verbio STT normalization


xquanluu:
Hi @davehorton 

Verbio was missing stt normalization that make jambonz webapps/recent call crash to show speech detail on record.

davehorton (approved):

PR #778, opened 2024-06-08
aws roleArn fix

vdharashive:
roleArn role need region as well

### DIFF
--- a/lib/tasks/stt-task.js
+++ b/lib/tasks/stt-task.js
@@ -196,8 +196,8 @@ class SttTask extends Task {
       credentials = {...credentials, access_token, stt_region};
     } else if (['aws', 'polly'].includes(vendor) && credentials.roleArn) {
       /* get aws access token */
-      const {roleArn} = credentials;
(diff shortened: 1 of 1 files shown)


xquanluu:
Thanks @vdharashive  a lot, I love the PR.

Hi @davehorton I'm good with this. we are using STS client do convert role arn to temporarily AWS Credential, STS client require to have region.

Not Sure how could we passed test with this bug !!!

davehorton:
please fix linting errors.

Also, not loving this bit of ugliness
```
const {accessKeyId, secretAccessKey, sessionToken, servedFromCache} = await getAwsAuthToken(null,null,region,roleArn);
```

davehorton:
I'd prefer to rewrite this
```js
async function getAwsAuthToken(
  logger, createHash, retrieveHash,
  awsAccessKeyId, awsSecretAccessKey, awsRegion, roleArn = null) {

```
as this
```js
async function getAwsAuthToken(
  logger, createHash, retrieveHash,
  {awsAccessKeyId, awsSecretAccessKey, awsRegion, roleArn}) {
```

same with getIbmAccessToken and getNuanceAccessToken

Issue #772, opened 2024-06-03
Buffering for deepgram is causing delayed transcripts


shanbhardwaj:
if `utteranceEndMs` or `endpointing` is set for deepgram then the finalized sentences are not received till the `UtteranceEnd` or `speech_final=true` marker is received at the feature server. This causes a delayed experience at the users end. 

This buffering here is causing us an issue. https://github.com/jambonz/jambonz-feature-server/blob/main/lib/tasks/gather.js#L929

davehorton:
This is by design, and the change suggested would indeed introduce a bug.  I suggest you add a `partialResultHook` to your gather in which case you should receive all partial results as they come and you can implement whatever logic you want

shanbhardwaj:
@davehorton Attached here are the logs for the following configuration 
```

transcribe: {
              transcriptionHook: '/transcribe',
              recognizer: {
                vendor: process.env.JAMBONZ_STT_VENDOR,
                language: 'en-US',
                separateRecognitionPerChannel: true,
                interim: true,
                deepgramOptions: {
                  model: 'nova-2',
                  punctuate: true,
                  redact: 'pci',
                  diarize: true,
                  numerals: true,
                  utteranceEndMs: 1000,
                  smartFormatting: true,
                },
              },
            },
```
The logs are here 
[freeswitch_logs.txt](https://github.com/user-attachments/files/15546249/freeswitch_logs.txt)
[feature_server_logs.txt](https://github.com/user-attachments/files/15546250/feature_server_logs.txt)

Following is the logs of the transcripts that were received when the logs were taken, which show the interim results are hard to use as they don't match the buffered transcript. 

The `is_final` node is always false for the interim results received from jambonz feature server.  It is set to true only when the buffered transcript is sent on utterance end. 

```
is_final false
Interim Transcript: If you know your party's
speech_final:  false

is_final false
Interim Transcript: If you know your party's extension, you *
speech_final:  false

is_final false
Interim Transcript: Press * for
speech_final:  false

is_final false
Interim Transcript: Press * for support.
speech_final:  false

is_final false
Interim Transcript: Press * for bill
speech_final:  false

is_final false
Interim Transcript: Press * to speak
speech_final:  false

is_final false
Interim Transcript: Press * to speak with our retention
speech_final:  false

is_final false
Interim Transcript: Press * if
speech_final:  false

is_final false
Interim Transcript: Press * if you are looking to start an
speech_final:  false

is_final false
[...]

PR #782, opened 2024-06-12
send end of utterance events if using deepgram, interim events are en…

davehorton:
…abled, and utterance_end_ms option is set (#772)


shanbhardwaj:
@davehorton this worked! 
We can close this issue

Issue #783, opened 2024-06-13
bargein with min words less than length of final transcript


davehorton:
Scenario:
 
You use a Set Session Node to set barge in with minimum words 2
You then ask a yes/no question
If you simply say “yes” (not barging in, but after the question), it is not triggered. You need to say more than 2 words.
 
It seems that the barge in minimum word settings are used even if we’re already done playing the prompt.

Issue #765, opened 2024-05-30
feat: actionHookDelayAction to be also triggered on no communication between jambonz and App server 


vdharashive:
The actionHookDelayAction is currently triggered only after the gather action is executed and while awaiting a response from the WSS app server. However, there are scenarios where there are no pending actions in the Jambonz queue. In these cases, the actionHookDelayAction should be triggered based on the specified configuration of actionHookDelayAction .

PR #785, opened 2024-06-14
fix for #765

davehorton:


davehorton:
actually fixes https://github.com/jambonz/jambonz-feature-server/issues/783

Issue #795, opened 2024-07-01
Errors on filler noises


paulotelles:
When using the filler noises there are two errors:

The first one is related to a race condition, where the filler starts within the gather, then we have an answer from the vendor and following we have a a filler noise for the waiting for new commands. This second filler noise does not works cause one was just playing and killed.

The second one is that the filler noises timeout does not work in the case " Waiting for new commands", it automatically starts playing and the delay does not work at all.

davehorton:
some of these seem like bugs that have been addressed.  Can you indicate which tag you are testing with that has these issues?

PR #796, opened 2024-07-01
fix errors on filler noises #795 795

paulotelles:
fix errors on filler noises #795 795

related to issue #795


lib/session/call-session.js
```diff
@@ -2504,10 +2506,26 @@ Duration=${duration} `
     }
   }
 
+  _startFillerNoiseTimer() {
+    this._clearFillerNoiseTimer();
+    this._fillerNoiseTimer = setTimeout(() => {
+      if (this.awaitCommands) {
```
Catharsis68, line 2512:
```suggestion
      if (this.awaitCommands && this.ep?.connected) {
```

Issue #777, opened 2024-06-07
feat: gather timer become obsolete  when any interm/hypostasis received 


vdharashive:
if gather has timeout of 10 sec and user start speaking at 7th sec, there are some interm/hypothesis received then gather should end at 10th sec . rather it should continue getting result until is-final =true is received , in current implimentation it getting terminated at 10 sec with timeout

davehorton:
yes, but in that case, when a partial transcript is received at 7th second, do you want it then to extend the timeout by another 10 seconds?

vdharashive:
timeout should be killed and final transcript has to come from asr , this logic only will apply if interim result are come

davehorton:
ok so once an interim transcript is received the timeout will be completely removed and we will wait for the final transcript.  In the case of continuous asr however that timeout will not be removed and will continue to be active

PR #800, opened 2024-07-05
clear gather timeout if imterim result received

xquanluu:
#777

### DIFF
--- a/lib/tasks/gather.js
+++ b/lib/tasks/gather.js
@@ -887,8 +887,8 @@ class TaskGather extends SttTask {
           this._sonioxTranscripts.push(evt.vendor.finalWords);
         }
       }
-      // If transcription received, clear timeout timer.
-      this._clearTimer();
+      // If transcription received, reset timeout timer.
+      this._startTimer();
       /* restart asr timer if we get a partial transcript */
       if (this.isContinuousAsr) this._startAsrTimer();
     }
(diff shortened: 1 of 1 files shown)


davehorton:
I have a concern with this change.  If we ever receive an interim transcript and then never get a final transcript, we will wait forever.  I believe I have seen this happen with google.  So I don't think I can merge this PR until it has been tested pretty extensively with google STT

xquanluu:
Hi @davehorton @vdharashive 

I agreed with Dave that gather timeout is a must and should not be clear to make sure gather will not be a deadlock. there are 2 options:
1/ reset gather timeout after interim transcript
2/ the application set higher value for gather timeout.

I would love to go with option 2 instead of this PR or option 1

vdharashive:
@xquanluu @davehorton we have tested this with google,deepgram and Microsoft no issues found , also i have made this live in production , client has mentioned the issue has been resolved. its safe to move to Merge this PR

vdharashive:
@xquanluu we can go with option  1/ reset gather timeout after interim transcript

davehorton:
> @xquanluu we can go with option 1/ reset gather timeout after interim transcript

I agree

vdharashive:
@xquanluu see if this can be completed as agreed

davehorton:
will merge after https://github.com/jambonz/jambonz-feature-server/pull/842

davehorton (approved):

PR #805, opened 2024-07-09
dtmf Barge-in disabled capturing input during prompt playing

vdharashive:
If dtmfbarge-in is disable and users provide input during the playing of prompt , the input is captured , rather it should start capturing input after playing the prompt

### DIFF
--- a/lib/tasks/gather.js
+++ b/lib/tasks/gather.js
@@ -201,7 +201,7 @@ class TaskGather extends SttTask {
           this.logger.debug('Gather: nested say task completed');
           this._stopVad();
           if (!this.killed) {
-            if (this.input.includes('digits') || this.asrDtmfTerminationDigit) {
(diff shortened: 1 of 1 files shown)


xquanluu:
Hi @vdharashive 

I just checked how speech bargein is working, by default, gather will start capturing user input during the prompt. if bargein = false, then gather still collect speech but does not stop the prompt.

The same thing is happening for dtmfBargein, this PR breaks the specification for dtmfBargein.

We have listenDuringPrompt properties, currently it's being used only for speech, if this flag listenDuringPrompt = false,jambonz will not capture dtmf during prompt. I'm creating new PR now.

vdharashive:
@xquanluu your suggest listenDuringPrompt = false the it will not capture dtmf and speech during the prompt is played/say ?

if this the case then i fine with listenDuringPrompt implimentation for dtmf as well

Issue #807, opened 2024-07-09
Introduce minConfidence property for gather verb


rammohanyadavalli:
The requirement is to return transcripts whose confidence levels are greater than or equal to the minConfidence property specified in the gather verb

davehorton:
on hold pending further review of this features

vdharashive:
gather notification should happen as "timeout" if confidence level is below its configured value and move to next verb

vdharashive:
@davehorton can you review this feature cc @rammohan-kore

PR #808, opened 2024-07-12
Check the confidence levels of a transcript  with minConfidence

rammohan-kore:
https://github.com/jambonz/jambonz-feature-server/issues/807
Check whether the confidence levels of a transcript is greater than or equal to the minConfidence property set in the gather verb, if confidence level of the transcript is less then resolve it as a timeout and proceed with next verb

### DIFF
--- a/lib/tasks/gather.js
+++ b/lib/tasks/gather.js
@@ -790,6 +790,8 @@ class TaskGather extends SttTask {
         if (!['soniox', 'aws', 'microsoft', 'deepgram'].includes(this.vendor)) this._startTranscribing(ep);
       }
       else {
+        /* this was removed to fix https://github.com/jambonz/jambonz-feature-server/issues/783 */
+        /*
         if (this.bargein && (words + bufferedWords) < this.minBargeinWordCount) {
           this.logger.debug({evt, words, bufferedWords},
             'TaskGather:_onTranscription - final transcript but < min barge words');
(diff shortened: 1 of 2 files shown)


davehorton (changes requested):
please refer to latest speech-utils, where minConfidence is a property on recognizer.  Also this feature should work both for gather and config (ie to set a session-level default).

lib/tasks/gather.js
```diff
@@ -736,6 +736,18 @@ class TaskGather extends SttTask {
...
+    const minConfidence = this.data?.recognizer?.minConfidence;
+    this.logger.debug({evt},
+      `TaskGather:_onTranscription - confidence (${confidence}), minConfidence (${minConfidence})`);
+    if (confidence && minConfidence && confidence < minConfidence) {
+      this.logger.info({evt},
+        'TaskGather:_onTranscription - rejecting transcript as confidence ' +
+        `(${confidence}) < minConfidence (${minConfidence})`);
+      evt.error = 'Rejecting transcript as confidence ' +
+        `(${confidence}) < minConfidence (${minConfidence})`
+      return this._resolve('stt-min-confidence-error', evt);
```
davehorton, line 749: could we rename this event as `stt-low-confidence`?  Also, I don't really consider this an "error", as nothing necessarily went wrong.

lib/tasks/gather.js
```diff
@@ -1091,6 +1103,12 @@ class TaskGather extends SttTask {
           this.emit('stt-error', evt);
           returnedVerbs = await this.performAction({reason: 'error', details: evt.error});
         }
+      } else if (reason.startsWith('stt-min-confidence-error')) {
+        if (this.parentTask) this.parentTask.emit('stt-min-confidence-error', evt);
+        else {
+          this.emit('stt-min-confidence-error', evt);
+          returnedVerbs = await this.performAction({reason: 'error', details: evt.error});
```
davehorton, line 1110: I do not think we should be sending a webhook with reason 'error'.  This is not an error -- errors are when things fail, this is simply a case of not getting very recognizable audio.

davehorton (changes requested):

lib/tasks/gather.js
rammohan-kore: Yes, renamed stt-min-confidence-error as stt-min-confidence

lib/tasks/gather.js
```diff
@@ -736,6 +736,16 @@ class TaskGather extends SttTask {
       this.logger.info({evt}, 'TaskGather:_onTranscription - got empty transcript, continue listening');
       return;
     }
+    const confidence = evt.alternatives[0].confidence;
+    const minConfidence = this.data?.recognizer?.minConfidence;
```
davehorton, line 740: minor point, but please remove ? after this.data - there is always a 'this.data' object

rammohan-kore: Remove ? after this.data

davehorton:
please rebase to latest main

davehorton:
also please run `npm run jslint` to make sure there are no linting errors

rammohan-kore:
Rebased with main branch and ran jslint

PR #817, opened 2024-07-19
fix config.transcribe should not override config.transcribe.recognizer

xquanluu:
Fix FD 288


xquanluu:
FD 288 already confirmed that PR is working for them

davehorton (approved):

PR #823, opened 2024-07-23
feat/813 - notify speech-bargein-detected and dtmf-bargein-detected events

rammohan-kore:
when the prompt message is in progress, send speech-bargein-detected or dtmf-bargein-detected verb:status

### DIFF
--- a/lib/tasks/gather.js
+++ b/lib/tasks/gather.js
@@ -321,8 +321,10 @@ class TaskGather extends SttTask {
     this.logger.debug(evt, 'TaskGather:_onDtmf');
     clearTimeout(this.interDigitTimer);
     let resolved = false;
-    if (this.dtmfBargein && !this.playComplete) {
-      this.notifyStatus({event: 'dtmf-bargein-detected', ...evt});
+    if (this.dtmfBargein) {
+      if (!this.playComplete) {
+        this.notifyStatus({event: 'dtmf-bargein-detected', ...evt});
+      }
       this._killAudio(cs);
       this.emit('dtmf', evt);
     }
--- a/lib/tasks/gather.js
+++ b/lib/tasks/gather.js
@@ -737,6 +737,16 @@ class TaskGather extends SttTask {
       this.logger.info({evt}, 'TaskGather:_onTranscription - got empty transcript, continue listening');
       return;
     }
+    const confidence = evt.alternatives[0].confidence;
+    const minConfidence = this.data.recognizer?.minConfidence;
+    this.logger.debug({evt},
+      `TaskGather:_onTranscription - confidence (${confidence}), minConfidence (${minConfidence})`);
+    if (confidence && minConfidence && confidence < minConfidence) {
+      this.logger.info({evt},
+        'TaskGather:_onTranscription - Transcript confidence ' +
+        `(${confidence}) < minConfidence (${minConfidence})`);
+      return this._resolve('stt-low-confidence', evt);
+    }
 
     /* fast path: our first partial transcript exactly matches an early hint */
     if (this.earlyHintsMatch && evt.is_final === false && this.partialTranscriptsCount++ === 0) {
@@ -1092,6 +1102,12 @@ class TaskGather extends SttTask {
         else {
           this.emit('stt-error', evt);
           returnedVerbs = await this.performAction({reason: 'error', details: evt.error});
+        }
+      } else if (reason.startsWith('stt-low-confidence')) {
+        if (this.parentTask) this.parentTask.emit('stt-low-confidence', evt);
+        else {
+          this.emit('stt-low-confidence', evt);
+          returnedVerbs = await this.performAction({reason: 'stt-low-confidence'});
         }
       }
     } catch (err) {  /*already logged error*/ }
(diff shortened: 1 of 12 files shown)


davehorton:
@rammohan-kore this needs to be rebased to latest main

rammohan-kore:
> @rammohan-kore this needs to be rebased to latest main

merged main branch into this feature branch

lib/tasks/gather.js
```diff
@@ -324,6 +324,7 @@ class TaskGather extends SttTask {
     if (this.dtmfBargein) {
       this._killAudio(cs);
       this.emit('dtmf', evt);
+      this.notifyStatus({event: 'dtmf-bargein-detected', ...evt});
```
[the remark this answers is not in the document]
rammohan-kore: When a prompt is being played, if user presses any keys on the dialpad on his phone, after the audio is  killed the dtmf-bargein-detected is sent once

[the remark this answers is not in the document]
rammohan-kore: Here is what I tested, when a prompt is being played and in between user presses a key, then the audio is getting killed and the dtmf-bargein-detected event is sent, however, if user presses other keys (after audio is killed), it is sending the dtmf-bargein-detected. I should be checking this.playComplete before sending ?

davehorton:
I would test this example:
- gather with nested say, input dtmf and numDigits: 4
- user enters 1, 2, 3, 4
- confirm dtmf-bargein-detected is sent once, only when dtmf "1" is detected

rammohan-kore:
> I would test this example:
> 
> * gather with nested say, input dtmf and numDigits: 4
> * user enters 1, 2, 3, 4
> * confirm dtmf-bargein-detected is sent once, only when dtmf "1" is detected

Yes @davehorton the event is sent multiple times, if I add if (!this.playComplete) it is sent only once

rammohan-kore:
Should we change  the if condition in _killAudio  from  if (this.hasFillerNoise || (!this.sayTask && !this.playTask && this.bargein))  to     if (this.hasFillerNoise || (!this.sayTask && !this.playTask && (this.bargein | this.dtmfBargin))) 

without this change the configuration, this.playComplete is not set and multiple digits are sent even after audio is killed

davehorton:
No, it seems to me that `if` statement is only triggered when there is neither a nested say nor play.  I would think instead in _killAudio we should add this final line at the end of the function
```
    if (this.playTask && !this.playTask.killed) {
      this.playTask.removeAllListeners('playDone');
      this.playTask.kill(cs);
      this.playTask = null;
    }
    this.playComplete = true;
```

rammohan-kore:
> No, it seems to me that `if` statement is only triggered when there is neither a nested say nor play. I would think instead in _killAudio we should add this final line at the end of the function
> 
> ```
>     if (this.playTask && !this.playTask.killed) {
>       this.playTask.removeAllListeners('playDone');
>       this.playTask.kill(cs);
>       this.playTask = null;
>     }
>     this.playComplete = true;
> ```

after adding this.playComplete = true at the end of _killAudio multiple digits are not sending dtmf-bargein-detected event

lib/tasks/gather.js
```diff
@@ -322,6 +322,9 @@ class TaskGather extends SttTask {
     clearTimeout(this.interDigitTimer);
     let resolved = false;
     if (this.dtmfBargein) {
+      if (!this.playComplete) {
```
davehorton, line 325: please change to a single condition
```
if (this.dtmfBargein && !this.playComplete) this.notifyStatus..
```

davehorton (changes requested):

lib/tasks/gather.js
rammohan-kore: added !this.playComplete to the main if condition

vdharashive:
@rammohan-kore can you complete this please

davehorton:
please rebase to main

rammohan-kore:
> please rebase to main

merged main into this feat/813 branch

lib/tasks/gather.js
```diff
@@ -321,7 +321,8 @@ class TaskGather extends SttTask {
     this.logger.debug(evt, 'TaskGather:_onDtmf');
     clearTimeout(this.interDigitTimer);
     let resolved = false;
-    if (this.dtmfBargein) {
+    if (this.dtmfBargein && !this.playComplete) {
```
davehorton, line 324: this doesnt look correct to me.  Given this code, once the play is complete we stop emitting dtmf events.  I understand you want to send a notify when the first dtmf is detected but you have included other side affects.

davehorton (changes requested):

lib/tasks/gather.js
rammohan-kore: if condition corrected and also referring the 0.2.9 version of time-series

davehorton (changes requested):
In this latest commit, it appears as if you manually modified package-lock.json ?  I see a change to time-series@0.2.9 in the dependencies section, but then down below I see references to time-series@.0.28.

What you need to do is update package.json, as you have done, and then run `npm install` to generate the package-lock.json.  You should never be editing the package-lock.json file directly

rammohan-kore:
> In this latest commit, it appears as if you manually modified package-lock.json ? I see a change to time-series@0.2.9 in the dependencies section, but then down below I see references to time-series@.0.28.
> 
> What you need to do is update package.json, as you have done, and then run `npm install` to generate the package-lock.json. You should never be editing the package-lock.json file directly

My bad, will ensure going forward, I have generated the package-lock.json using npm -i

davehorton (approved):

PR #825, opened 2024-07-24
fix conference end is not sent when moderator leave conference

xquanluu:


lib/session/call-session.js
```diff
@@ -2067,6 +2067,11 @@ Duration=${duration} `
       this.logger.error('CallSession:replaceEndpoint cannot be called without stable dlg');
       return;
     }
+    // When this call kicked out from conference, session need to replace endpoint
+    // but this.ms might be undefined/null at this case.
+    if (!this.ms) {
```
davehorton, line 2072: can we just have
```
this.ms = this.ms || this.getMS();
```

xquanluu: fixed, the this.getMS() already done that.

this.ms = this.getMS();

lib/session/call-session.js
```diff
@@ -2069,9 +2069,7 @@ Duration=${duration} `
     }
     // When this call kicked out from conference, session need to replace endpoint
     // but this.ms might be undefined/null at this case.
-    if (!this.ms) {
-      this.ms = this.getMS();
-    }
+    this.ms = this.getMS();
```
davehorton, line 2072: not what I asked for

davehorton (changes requested):

lib/session/call-session.js
xquanluu: But

```
  /**
   * get a media server to use for this call
   */
  getMS() {
    if (!this.ms) {
      this.ms = this.srf.locals.getFreeswitch();
      if (!this.ms) {
        this._mediaServerFailure = true;
        throw new Error('no available freeswitch');
      }
    }
    return this.ms;
  }
```

so that this.ms = this.getMS(); is already enough

davehorton: I prefer the other because although it is functionally equivalent it signals clearly to the user that this is not needed in all cases, only in a special case

xquanluu: I see, fixed

davehorton (approved):

vdharashive:
@xquanluu its working as expected please merge it

davehorton (approved):

PR #829, opened 2024-07-25
fixed cannot get application tts/stt speech credentials if there is n…

xquanluu:
…o verb's synthesizer/recognizer

Jambonz has 2 levels for synthesizer/recognizer: in application and inside a verb.

Currently, the label is following only verb's synthezier/recognizer, If verb does not provide any synthezier/recognizer, only vendor, language will use default application level but label does not use application's synethsizer/recognizer which is wrong.


lib/tasks/stt-task.js
```diff
@@ -44,6 +44,9 @@ class SttTask extends Task {
       }
     } else {
       this.data.recognizer = {hints: [], altLanguages: []};
+      // If there is no verb's recognizer, use application's recognizer for label
```
davehorton, line 47: hmm, whoa.....default to a label named 'default' ?  Why is this necessary?  What if the user has a label actually named 'default'?

vdharashive:
@xquanluu @davehorton if user set label then only its value need to be considered and it should not use any default from application level if its not set , last PR for label also failed some the test cases i am ok to revert that https://github.com/jambonz/jambonz-feature-server/commit/5ccea65b7f0e28ae58cf43dada40a5d6f53bec8e

lib/tasks/stt-task.js
xquanluu: the logic later will check if label === default, then use cs.syntheizerLabel

We need to have a condition to identify when to use application default label.

xquanluu:
by default, without any verb synthezier or recognizer, we have to use application.systhesizer/recognizer label

lib/tasks/stt-task.js
xquanluu: By using this special string: {jambonz_label}

If there is no verb synthesizer/recognizer, jambonz will replace this special string to application synthesizer/recognizer label.

davehorton:
@xquanluu can you rebase?  Also we should discuss the overall approach (the special case logic around a "magic label")

vdharashive:
@xquanluu with google asr fallback is failing using config

Issue #822, opened 2024-07-23
No Call Events if caller hangs up before answer


sammachin:
If I have an incomming call where only Early Media is played to the caller (eg the call is not answered) and the caller hangs up I do not recieve any call status events to indicate this, only the initial trying event before the EM is played.

Similarly if the last action in the JSON is the play of EM the call is not ended by jambonz (and no events are recieved) even if I add a hangup verb after the play.

Tested on jambonz.cloud 23rd July, (v0.9.1?)

SIP_call_id 2024072312022200013@2900-0601-0284-86
at 2024-07-23T11:02:22.508Z
The call_sid field is null in the portal

davehorton:
@surajshivkumar please reference this issue when you have a PR ready to review

PR #831, opened 2024-07-25
Call Events - Early media call status

surajshivkumar:


davehorton:
generally we want all PRs to reference an issue.  In this case we have not created an issue yet, but please create an issue describing the problem

sammachin:
While this looks like a useful event to be sending its not really solving the issue I raised as the event I'm interested in is where the caller hangs up the call during early media

xquanluu:
Hi @sammachin 

while testing this issue, we found that if the call is hangup by caller, jambonz will sent event NoAnswer, sip reason is Request Terminated.

This is correct behavior, please let us know if you was not received this event. If no, please attach a log here.

lib/session/inbound-call-session.js
```diff
@@ -33,13 +33,24 @@ class InboundCallSession extends CallSession {
...
+    this.rootSpan.setAttributes({'call.termination': 'caller abandoned'});
     this.rootSpan.setAttributes({'call.termination': 'caller abandoned'});
     this.callInfo.callTerminationBy = 'caller';
+    const wasEarlyMedia = this.callInfo.callStatus === 'early-media';
     this._notifyCallStatusChange({
       callStatus: CallStatus.NoAnswer,
       sipStatus: 487,
       sipReason: 'Request Terminated'
     });
+    if (wasEarlyMedia) {
```
xquanluu, line 45: I don't think we need this event. Please consider to remove it.

NoAnswer is already good enoug as it's final status as same as Completed

lib/session/inbound-call-session.js
```diff
@@ -33,13 +33,24 @@ class InboundCallSession extends CallSession {
   }
 
   _onCancel() {
+    this.rootSpan.setAttributes({'call.termination': 'caller abandoned'});
```
xquanluu, line 36: Could we consider to remove this duplication.

surajshivkumar: Sure. We can remove this.

sammachin:
> Hi @sammachin
> 
> while testing this issue, we found that if the call is hangup by caller, jambonz will sent event NoAnswer, sip reason is Request Terminated.
> 
> This is correct behavior, please let us know if you was not received this event. If no, please attach a log here.

Thats what I would have expected but not the behcaviour I saw, have a look at #822 for details of a call on jambonz.cloud with the issue.

surajshivkumar:
@sammachin Could you kindly supply us with another example, or reproduce the same example and provide us with the callsid where you are experiencing the problem?

sammachin:
Sure, call sid 7b2165d5-2f56-4553-8472-620adea9980f at 08:08:49 UTC on 2nd August

PR #835, opened 2024-07-31
Bug/fix missing arg reconnect alert

Catharsis68:
Fix a bug where the argument `url` was missing when writing an alert after receiving a `reconnect-error`

Screenshot from 2024-07-31 10-18-42

Fix vulnerabilities via `npm audit fix`

Before: 20 vulnerabilities (4 moderate, 16 high)
After: 4 vulnerabilities (3 moderate, 1 high)

### DIFF
--- a/lib/session/call-session.js
+++ b/lib/session/call-session.js
@@ -1806,7 +1806,8 @@ Duration=${duration} `
       await writeAlerts({
         alert_type: AlertType.WEBHOOK_CONNECTION_FAILURE,
         account_sid: this.accountSid,
-        detail: `Session:reconnect error ${err}`
+        detail: `Session:reconnect error ${err}`,
+        url: this.application.call_hook.url,
       });
     } catch (error) {
       this.logger.error({error}, 'Error writing WEBHOOK_CONNECTION_FAILURE alert');
--- a/package-lock.json
+++ b/package-lock.json
@@ -79,1460 +79,687 @@
         "node": ">=6.0.0"
       }
     },
-    "node_modules/@aws-crypto/ie11-detection": {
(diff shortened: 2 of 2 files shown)


lib/session/call-session.js
```diff
@@ -1806,7 +1806,8 @@ Duration=${duration} `
       await writeAlerts({
         alert_type: AlertType.WEBHOOK_CONNECTION_FAILURE,
         account_sid: this.accountSid,
-        detail: `Session:reconnect error ${err}`
+        detail: `Session:reconnect error ${err}`,
+        url: this.application.call_hook.url,
```
davehorton, line 1810: you can not simply add `url` to the alert if it is not part of the schema.  You should first add a field to the schema for url (in time-series)

davehorton (changes requested):

lib/session/call-session.js
davehorton: Oh sorry, I see you just want it logged, not added to the schema. Then this looks ok

package-lock.json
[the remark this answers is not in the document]
Catharsis68: Because I run `npm audit fix`

Before: 20 vulnerabilities (4 moderate, 16 high)
After: 4 vulnerabilities (3 moderate, 1 high)

lib/session/call-session.js
Catharsis68: In other places where the same alert is created, the url is provisioned. Only in the "reconnection" case the `url` is missing

Screenshot from 2024-07-31 15-06-33

davehorton (approved):

Issue #836, opened 2024-07-31
Capture vendor and target_sid (call_sid) during stt/tts failure  alerts 


rammohan-kore:
The alert mechanism need to capture vendor, call_sid (target_sid) whenever a STT, TTS alerts are raised

PR #838, opened 2024-07-31
feat/836: capturing callSid for STT and TTS alerts

rammohan-kore:
https://github.com/jambonz/jambonz-feature-server/issues/836

Sending callSid for STT/TTS alerts

### DIFF
--- a/lib/tasks/say.js
+++ b/lib/tasks/say.js
@@ -114,7 +114,8 @@ class TaskSay extends TtsTask {
         writeAlerts({
           account_sid,
           alert_type: AlertType.TTS_NOT_PROVISIONED,
-          vendor
+          vendor,
+          target_sid: cs.callSid
         }).catch((err) => this.logger.info({err}, 'Error generating alert for no tts'));
         throw new Error('no provisioned speech credentials for TTS');
       }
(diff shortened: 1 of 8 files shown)


lib/tasks/say.js
```diff
@@ -201,7 +203,7 @@ class TaskSay extends TtsTask {
   }
 
   async exec(cs, {ep}) {
-    const {srf, accountSid:account_sid} = cs;
+    const {srf, accountSid:account_sid, target_sid:callSid} = cs;
```
davehorton, line 206: this actually looks backwards.  You are trying to extract `target_sid` from `cs`, which does not exist, and assign it to callSid.  This tells me also that you have not tested this properly.  Please test and verify the data is being written to the time-series database

davehorton (changes requested):

lib/tasks/say.js
rammohan-kore: My bad, apologies for this, as I was not able to generate this scenario in the system it got missed, I have corrected it now, also added missing entries for STT_FAILURE alerts

rammohan-kore:
> please include this change in your PR and I will merge:
> 
> [e33c0ce](https://github.com/jambonz/jambonz-feature-server/commit/e33c0ce9911110befba17173dfbf8e1f349b3e48)

Included the above commit

davehorton (approved):

Issue #758, opened 2024-05-22
Alternative Google Speech langauge detection normalisation 


vdharashive:
Google speech detection format is not same as Microsoft below are sample of both 

detected language is not coming in speech language_code  parameter for google as its as its comming in microsoft , both the format need to same  

Google 
"speech": {
            **"language_code": "en-US",**
            "channel_tag": 1,
            "is_final": true,
          "alternatives": [
                {
                    "confidence": 0.9832135438919067,
                    "transcript": "मेरा नाम क्या है"
                }
            ],

Microsoft 
"speech": {
                **"language_code": "fr-CA",**
                "channel_tag": 1,
                "is_final": true,
                "alternatives": [
                    {
                        "confidence": 0.7186232,
                        "transcript": "Juste en dessous de 10000"
                    }
                ],

===============================================
Google Speech Detection 

"speech": {
            "language_code": "en-US",
            "channel_tag": 1,
            "is_final": true,
            "alternatives": [
                {
                    "confidence": 0.9832135438919067,
                    "transcript": "मेरा नाम क्या है"
                }
            ],
            "vendor": {
                "name": "google",
                "evt": {
                    "stability": 0,
                    "is_final": true,
                    "alternatives": [
                        {
                            "confidence": 0.9832135438919067,
                            "transcript": "मेरा नाम क्या है"
                        }
                    ],
                    "language_code": "hi-in",
                    "channel_tag": 0,
                    "result_end_time": 5550
                }
            }
        },
        "reason": "speechDetected"
        
        
        
        

Microsoft Speech detection

            "speech": {
                "language_code": "fr-CA",
                "channel_tag": 1,
                "is_final": true,
                "alternatives": [
                    {
                        "confidence": 0.7186232,
                        "transcript": "Juste en dessous de 10000"
                    }
                ],
                "vendor": {
                    "name": "microsoft",
                    "evt": {
                        "id": "3f9de508c68e40a6896bb37ae2a92dc2",
                        "recognition_status": "Success",
                        "offset": 42700000,
                        "duration": 17200000,
                        "primary_language": {
                            "language": "fr-CA",
                            "confidence": "Low"
                        },
                        "channel": 0,
                        "display_text": "Juste en dessous de 10000.",
                        "n_best": [
                            {
                                "confidence": 0.7186232,
                                "lexical": "juste en dessous de dix mille",
                                "itn": "juste en dessous de 10000",
                                "masked_i_t_n": "juste en dessous de 10000",
[...]

PR #843, opened 2024-08-06
Feat/758

rammohan-kore:
in transcription-utils normalizeGoogle method setting the language_code taken from evt object


davehorton:
@rammohan-kore in future please put a reference to related issue in the PR commit notice.  In this case, that would be #758

davehorton (approved):

PR #855, opened 2024-08-12
fix support precache audio with tts stream

xquanluu:


xquanluu:
wait for speech util new release

xquanluu:
FD_320 already confirmed the issue has been fixed for tts precache together with tts streaming

davehorton (approved):

Issue #868, opened 2024-08-16
Use the properties from global config in verb for TTS


rammohanyadavalli:
The say verb is not using the options specified in global config

vdharashive:
Following test cases need to be test with 

1. using config set synthesizer with option  
2.  say have synthesizer with option 
3.  gather with config synthesizer with option
4. gather with say having synthesizer with option

davehorton:
@rammohanyadavalli please confirm when you have tested above test cases

rammohan-y:
> @rammohanyadavalli please confirm when you have tested above test cases

Yeah, we have tested the above scenarios

PR #869, opened 2024-08-16
Feat/868: Use global synthesizer config  properties for say verb

rammohan-y:
https://github.com/jambonz/jambonz-feature-server/issues/868

Use the global synthenizer config from config verb for say verb


lib/tasks/say.js
```diff
@@ -210,7 +211,22 @@ class TaskSay extends TtsTask {
 
     await super.exec(cs);
     this.ep = ep;
+    if (cs.synthesizer) {
```
davehorton, line 214: shouldnt this be done in the superclass as well?

davehorton (changes requested):

lib/tasks/say.js
xquanluu: Hi @davehorton  @rammohan-y 

I see that _synthesizeWithSpecificVendor is implemented in 2 places tts-task and say task and they are doing exactly the same excep som logging and notification that tts-task doing better.

I would suggest to migrate the _synthesizeWithSpecificVendor in say.js to tts-task and we have only one place to handle thing.

davehorton: @rammohan-y I agree with @xquanluu's point above, can you make that change?

rammohan-y: @davehorton moved _synthesizeWithSpecificVendor method from say.js to tts-task.js

rammohan-y: > Thanks, please do a thorough regression test, as this is a big change

The following were testing for both say and gather  + say

config contains synthesizer config, say verb doesn't contain synthesizer: synthesizer config from config is taken
config does not contain sythesizer config, verb contains synthesizer: synthesizer config from verb is taken
both config and verb has synthesizer options: synthesizer config from verb is taken
both config and verb does not contain synthesizer options: system took what is configured in the application

rammohan-y: Tested for playht, microsoft
Tested with disableTtsCache true

davehorton:
can rebase @rammohan-y

rammohanyadavalli:
> can rebase @rammohan-y

Yeah, rebased now @davehorton

davehorton (approved):

PR #875, opened 2024-08-22
Add support for configuring the IP address that is advertised to the API Server

zscgeek:
We were running into a problem where if the feature server was running on AWS fargate ECS it would get the wrong ip address returned. The issue is the first network card in a fargate container is amazon's metadata network. The result is that the node `ip.address()` function would return that IP, instead of the IP that is on the VPC. This change allows us to override that value that is registered in redis and used by the API server to reach the feature server.


zscgeek:
the issue for this is https://github.com/jambonz/jambonz-feature-server/issues/876

davehorton:
please rebase to latest main

davehorton:
@zscgeek need you to rebase to latest

zscgeek:
rebased

davehorton (approved):

PR #878, opened 2024-08-23
Fix/audio issue kick conference

xquanluu:


xquanluu:
@vdharashive confirmed this is fixed, audio is good after kicked out from conference

vdharashive:
@xquanluu @davehorton this fix has fixed the RTP issue , this can be merged

davehorton:
@xquanluu is there more work to be done on this one?

xquanluu:
No Dave, please help to merge this.

lib/session/call-session.js
```diff
@@ -2074,6 +2074,9 @@ Duration=${duration} `
     // When this call kicked out from conference, session need to replace endpoint
     // but this.ms might be undefined/null at this case.
     this.ms = this.ms || this.getMS();
+    // Destroy previous ep if it's still running.
+    if (this.ep && this.ep.connected) this.ep.destroy();
```
davehorton, line 2078: can we change this to 
```
if (this.ep?.connected) this.ep.destroy()
```

davehorton (changes requested):

lib/session/call-session.js
xquanluu: fixed

PR #886, opened 2024-08-30
fix #883 that after kicked from conference, no long receive freeswitch CUSTOM event

### LINKED ISSUE #883: bug: after kicking the user from conference dial verb not sending transcript
after implementation of this fix 

https://github.com/jambonz/jambonz-feature-server/commit/4d077e990f95331ba59c281c576037baafd5a419

the kicked user from conference  on his channel execute dial verb with transcribe then the transcript for a-party are not coming but for b-party transcript are generating

xquanluu:
#883

### DIFF
--- a/lib/tasks/conference.js
+++ b/lib/tasks/conference.js
@@ -118,10 +118,9 @@ class Conference extends Task {
     this.emitter.emit('kill');
     await this._doFinalMemberCheck(cs);
     if (this.ep && this.ep.connected) {
-      // IMPORTANT!!!! Remove CUSTOM event will make the ep is longer usable for any orther activity
-      // all freeswitch modules are firing CUSTOM events
-      // There is issue after caller is kicked from conferrence, he cannot use gather/transcribe
-      // this.ep.conn.removeAllListeners('esl::event::CUSTOM::*');
+      // drachtio-fsmrf override esl::event::CUSTOM to conference join listerner, After finish the conference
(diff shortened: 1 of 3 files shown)


lib/tasks/conference.js
```diff
@@ -118,7 +118,10 @@ class Conference extends Task {
     this.emitter.emit('kill');
     await this._doFinalMemberCheck(cs);
     if (this.ep && this.ep.connected) {
-      this.ep.conn.removeAllListeners('esl::event::CUSTOM::*');
+      // IMPORTANT!!!! Remove CUSTOM event will make the ep is longer usable for any orther activity
+      // all freeswitch modules are firing CUSTOM events
+      // There is issue after caller is kicked from conferrence, he cannot use gather/transcribe
+      // this.ep.conn.removeAllListeners('esl::event::CUSTOM::*');
```
davehorton, line 124: perhaps we need to study this further, and figure out how to remove only event listeners for conference events, because I am pretty sure this code was in there for a reason (ie removing it may surface old bugs)

xquanluu: You are right.

I have created new PR in fsmrf: https://github.com/drachtio/drachtio-fsmrf/pull/99

that allow us to reset the esl CUSTOM event instead of completely remove it.

xquanluu: fixed, tested new change on jambonz.me and it's working

davehorton (approved):

PR #890, opened 2024-09-04
add support for aws language model name when transcribing

davehorton:


davehorton:
This change assumes that we are using the AWS websocket api to their transcribe service.  We do intend to switch from their grpc to the websocket api, but we must make the feature server backwards compatible such that it can detect that the older grpc api is being used and parse the payloads accordingly, because the payloads are different.  This will give us the ability to update to the websocket api in production yet fallback if we need to for some reason.

[Here is a comparison of the different payloads per api](https://gist.github.com/davehorton/2144d18ae50056b7569d89d98ea929be).  We can use this to determine which interface is being used and parse accordingly

vdharashive:
@davehorton please rebase this

davehorton:
@vdharashive rebased

Issue #891, opened 2024-09-09
custom-stt options are not sent to custom-stt's websocket


rammohan-y:
When setting options under customOptions.options object for custom-stt, the options are not sent to custom-stt's websocket as part of start event

"recognizer": {
                    "vendor":"custom:customvoice",
                    "customOptions": {
                        "options" : {
                            "foo" : "bar"
                        }
                    }
                  }

the options object under customOptions should be sent to the custom speech's websocket under start event's payload along with callSid

PR #892, opened 2024-09-09
feat/891: getting the options from customOptions in case of custom stt

rammohan-y:
https://github.com/jambonz/jambonz-feature-server/issues/891
getting the options from customOptions in case of custom-stt

### DIFF
--- a/lib/utils/transcription-utils.js
+++ b/lib/utils/transcription-utils.js
@@ -856,7 +856,7 @@ module.exports = (logger) => {
           {VERBIO_SPEECH_INCOMPLETE_TIMEOUT: verbioOptions.speech_incomplete_timeout}),
       };
     } else if (vendor.startsWith('custom:')) {
-      let {options = {}} = rOpts;
+      let {options = {}} = rOpts.customOptions || {};
       const {sampleRate} = rOpts.customOptions || {};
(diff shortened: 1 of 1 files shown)


xquanluu:
Based on this spec: https://github.com/jambonz/verb-specifications/blob/main/specs.json#L679

this PR is correct, Look good to me.

lib/utils/transcription-utils.js
```diff
@@ -856,7 +856,7 @@ module.exports = (logger) => {
           {VERBIO_SPEECH_INCOMPLETE_TIMEOUT: verbioOptions.speech_incomplete_timeout}),
       };
     } else if (vendor.startsWith('custom:')) {
-      let {options = {}} = rOpts;
+      let {options = {}} = rOpts.customOptions || {};
```
davehorton, line 859: shouldnt this simply be
```js
let options = rOpt.customOptions || {};
```

davehorton (changes requested):

lib/utils/transcription-utils.js
xquanluu: inside customOptions, there is options attribute. and we want to use that options.

davehorton (approved):

Issue #903, opened 2024-09-18
enhancement: start vad when config.vad.enable is true


rammohan-y:
currently vad detection happens only when gather verb's minBargeinWordCount is 0, we have vad at config verb, we should start vad detection when config.vad.enable is true

vdharashive:
The purpose of this is to start streaming audio to asr when vad is detected specially with Microsoft,  which send garbage Yes or No if it get empty audio stream

PR #905, opened 2024-09-18
feat/903: enabling vad when config.vad.enable is true

rammohan-y:
feat/903: enabling vad when config.vad.enable is true

https://github.com/jambonz/jambonz-feature-server/issues/903

### DIFF
--- a/lib/tasks/gather.js
+++ b/lib/tasks/gather.js
@@ -44,7 +44,6 @@ class TaskGather extends SttTask {
     this.interim = !!this.partialResultHook || this.bargein || (this.timeout > 0);
     this.listenDuringPrompt = this.data.listenDuringPrompt === false ? false : true;
     this.minBargeinWordCount = this.data.minBargeinWordCount !== undefined ? this.data.minBargeinWordCount : 1;
-    this._vadEnabled = this.minBargeinWordCount === 0;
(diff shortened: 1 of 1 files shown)


vdharashive:
@davehorton can you pls review it

davehorton:
@rammohan-y @vdharashive 
I am not sure what you are trying to do here, and I think we are confusing two different vad-related features.

1.  Recently, you spoke of wanting to delay the ASR connection for microsoft until vad is detected.  We used to have a feature like that built into the mod_azure_transcribe and [you can see remnants of it here](https://github.com/jambonz/freeswitch-modules/blob/8a3c001b5949f2f2ba4d8d1fbbd4e697fd2b544f/mod_azure_transcribe/azure_transcribe_glue.cpp#L562).  The code that set that env var has been removed from the feature server though (the feature was thought to be problematic in terms of creating support issues if we connected "too late" to catch some speech.  So this feature, which is what I thought you have been talking about recently, is not currently available and your changes in this PR do not resurrect it.
2.  If someone wants bargein to happen even faster than before we get a partial transcript back with one word back, they can specify minBargeinWords = 0 and then we use vad to detect speech energy barge in quickly.

Now, with that explanation, what exactly are you hoping to accomplish with this PR

PR #907, opened 2024-09-19
fixed adulting call session does not send status callback if hangup is used

xquanluu:


vdharashive:
@xquanluu this is working as expected , this can be merged @davehorton

davehorton (approved):

PR #912, opened 2024-09-24
fix for https://github.com/jambonz/freeswitch-modules/issues/117

davehorton:


xquanluu:
Hi @davehorton 

There is usecase that the primary TTS vendor is failed and jambonz did failover to 2nd vendor.

In next run we should keep using the 2nd vendor instead of retry again the primary.

Issue #913, opened 2024-09-25
Cannot use gather digits without nested say or play


xquanluu:
If Gather input does not have nested say or play, no dtmf is captured

PR #914, opened 2024-09-25
Fixed Gather digits does not work without nested say/play

xquanluu:
#913

### DIFF
--- a/lib/tasks/gather.js
+++ b/lib/tasks/gather.js
@@ -177,8 +177,10 @@ class TaskGather extends SttTask {
     this._startVad();
 
     const startDtmfListener = () => {
+      if (this._dtmfListenerStarted) return;
       if (this.input.includes('digits') || this.dtmfBargein || this.asrDtmfTerminationDigit) {
         ep.on('dtmf', this._onDtmf.bind(this, cs, ep));
+        this._dtmfListenerStarted = true;
       }
     };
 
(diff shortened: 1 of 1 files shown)


lib/tasks/gather.js
```diff
@@ -177,8 +177,10 @@ class TaskGather extends SttTask {
     this._startVad();
 
     const startDtmfListener = () => {
+      if (this._dtmfListenerStarted) return;
```
davehorton, line 180: Is there a scenario where this gets called twice?

xquanluu: You are right, we don't have a case that this startDtmfListener is called twice.

davehorton: In that case can you make it an assert instead so if it ever does happen we get an error we can fix
```
assert(! this._dtmfListenerStarted);
```

lib/tasks/gather.js
```diff
@@ -177,10 +177,8 @@ class TaskGather extends SttTask {
     this._startVad();
 
     const startDtmfListener = () => {
-      if (this._dtmfListenerStarted) return;
```
davehorton: if we are confident we are not going to call this twice, then I would still like to see an assert here

davehorton (changes requested):

lib/tasks/gather.js
xquanluu: I see
fixed

davehorton (approved):

Issue #911, opened 2024-09-24
feat: config to support refer hook 


vdharashive:
its some time needed to set refer hook on incoming call , by using config:refer verb hook can be set to handle the refer

PR #915, opened 2024-09-25
support config referHook

xquanluu:
#911

### DIFF
--- a/lib/session/call-session.js
+++ b/lib/session/call-session.js
@@ -20,6 +20,7 @@ const listTaskNames = require('../utils/summarize-tasks');
 const HttpRequestor = require('../utils/http-requestor');
 const WsRequestor = require('../utils/ws-requestor');
 const ActionHookDelayProcessor = require('../utils/action-hook-delay');
+const {parseUri} = require('drachtio-srf');
 const {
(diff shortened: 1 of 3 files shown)


lib/session/call-session.js
```diff
@@ -2232,17 +2245,59 @@ Duration=${duration} `
...
    */
   _onRefer(req, res) {
     const task = this.currentTask;
     const sd = task.sd;
-    if (task && TaskName.Dial === task.name && sd) {
+    if (task && TaskName.Dial === task.name && sd && task.referHook) {
       task.handleRefer(this, req, res);
     }
     else {
+      this._handleRefer(req, res);
```
davehorton, line 2259: this almost entirely duplicates code that is in the Dial task.  Can it be refactored into a common utils file?

davehorton (changes requested):

lib/session/call-session.js
xquanluu: They are not identical. the handleRefer from dial verb is handling for both legs which is totally different from call-session handle for incoming call leg only.

davehorton (approved):

Issue #924, opened 2024-10-04
Make actions as optional property when there is no noResponseTimeout in ActionHookDelayProcessor


rammohan-y:
Make actions as optional property when there is no noResponseTimeout in ActionHookDelayProcessor, as we might want to configure only giveUpActions

This was supposed to be part of https://github.com/jambonz/jambonz-feature-server/issues/902 but missed this feature

Issue #902, opened 2024-09-15
enhancement:  actionHookDelayAction to have action to be triggered for noResponseGiveupTimeout


vdharashive:
current implementation on triggering noResponseGiveupTimeout jambonz hangup the call , rather then hangup call there can be a new giveactions array which will have array of  verbs that needs to be executed in sequence 

{
  "verb": "config",
  "actionHookDelayAction" : {
    "enabled": true,
    "retries": 2,
    "noResponseTimeout": 5,
    "noResponseGiveupTimeout": 20,
    "actions": [
      {
        "verb": "say",
        "text": "Please hold while we complete your transaction"
      },
      {
        "verb": "say",
        "text": "Please continue to hold."
      }
    ],
  "giveupactions": [
      {
        "verb": "say",
        "text": "Thank you"
      },
      {
        "verb": "hangup",
      }
    ]

  }
}

giveupactions is optional configuration if giveupactions is not provided then jambonz can drop the call

PR #925, opened 2024-10-04
feat/924: made actions as optional when there is no noResponseTimeout

rammohan-y:
https://github.com/jambonz/jambonz-feature-server/issues/924

Making actions property as optional when noResponseTimeout is not specified or zero, as we have introduced noResponseGiveUpTimeout and giveUpActions, the noResponseTimout & actions array can be made optional

This is an extension to the feature https://github.com/jambonz/jambonz-feature-server/issues/902


vdharashive:
this validation was missing , can you merge this to be part of 0.9.2 release @davehorton

davehorton (approved):

Issue #929, opened 2024-10-11
Feat: Original sip Call-id to be part of call:status event


vdharashive:
currently there is no way to obtain original call-id from call:status event , it would be useful feature to track at bot level 

this needs to be there for inbound /outbound and restdial calls

vdharashive:
@xquanluu also can it be logged in recentcall cdr and  redis

vdharashive:
thanks @xquanluu

PR #934, opened 2024-10-14
support sbcCallId in calling/status hook

xquanluu:
https://github.com/jambonz/jambonz-feature-server/issues/929


lib/http-routes/api/create-call.js
```diff
@@ -293,6 +293,8 @@ router.post('/',
           },
           cbProvisional: (prov) => {
             const callStatus = prov.body ? CallStatus.EarlyMedia : CallStatus.Ringing;
+            // Update call-id for sbc outbound INVITE
+            cs.callInfo.sipCallid = prov.get('X-CID');
```
davehorton, line 297: can we rename this `sbcCallId` everywhere

xquanluu: Fixed to use sbcCallid

davehorton (approved):

Issue #935, opened 2024-10-15
Improvement : Oubound Register User call to be sent via registered sbc


vdharashive:
current implementation for outbound call for register user , feature server send call randomly to any available sbc-sip server , then sbc-sip server divert the call to the sbc-sip server where User endpoint is connected/registered , rather then doing this to check at feature server that which sbc user is connected and send call to that sbc-sip server . this will help reducing the Hops

PR #949, opened 2024-10-28
feature server should send USER call to the sbc sip that is connect with the user

xquanluu:
#935


vdharashive:
@xquanluu now all calls are going to registered user sbc 

We can merge the code

lib/utils/network.js
```diff
@@ -0,0 +1,28 @@
+
+/**
+ *
+ * @param {*} logger
+ * @param {*} hostport
+ * @param {*} protocol
+ * @returns array with bellow properties
```
davehorton, line 7: please fix typo "below"

lib/utils/network.js
```diff
@@ -0,0 +1,28 @@
...
+ * 0: protocol
+ * 1: ip address
+ * 2: port
```
davehorton, line 10: its actually not clear from this description what this function does. It seems to be filtering so as to return the ip address that the feature server is listening on for a given protocol.  Can you make the description state the purpose more clearly?

lib/http-routes/api/create-call.js
```diff
@@ -140,6 +141,16 @@ router.post('/',
         }
       }
 
+      // find handling sbc sip for called user
+      if (target.type === 'user') {
+        const { registrar }  = srf.locals.dbHelpers;
+        const reg = await registrar.query(target.name);
+        if (reg) {
+          sbcAddress = selectHostPort(logger, reg.sbcAddress, 'tcp')[1];
```
davehorton, line 149: Hmm, I am not loving this.  Feature server already selects an SBC to send to (prior to this change) and I believe it finds the IP address of the SBC in the VPC (the private address).  How is that done?  If different than this, why?

davehorton (commented):
I made some comments and additionally I think this feature should be enabled only when a env is set, something like JAMBONES_DIAL_SBC_FOR_REGISTERED_USER

lib/utils/network.js
xquanluu: fixed

lib/utils/network.js
xquanluu: fixed

lib/http-routes/api/create-call.js
xquanluu: @davehorton 

feature server look at redis for keys default:active-sip to get list of sbc private ip address.

then select next sbc address from the list to send request to. this way, feature server cannot send INVITE request to correct sbc which is handling for the user.

this code will get registration from redis for the user, get sbc_address which is handling for the user and as you know that there is a tcp interface from drachtio server will contains private IP address for the sbc. feature server use this private IP to send the request to that SBC.

davehorton: its ugly but for now I'll accept.  We need to come back and fix this at some point -- simplify the learning of sbc private/internal address

davehorton (approved):

PR #958, opened 2024-11-04
fixed dial verb should use calling id from From header

xquanluu:


lib/tasks/dial.js
```diff
@@ -507,10 +507,17 @@ class TaskDial extends Task {
       ...this.headers
     };
 
+    // get calling user from From header
+    const fromHeader = req.get('From');
+    const fromUser = fromHeader.match(/^\"(.+)\"\s*<sips?:.+@/);
```
davehorton, line 512: can you use the parsing functions in drachtio-srf?  ie
```
const parsed = req.getParsedHeader('From')
//... then check parsed.user

davehorton (approved):

xquanluu:
Hi @davehorton 

Tested on jambonz.me, it's working.

PR #969, opened 2024-11-12
Add support to export to more than one otel platform.

zscgeek:
This is helpful for if you want to keep using the bundled jaeger support in the web console AND send to external OTLP based platform (such as Axiom.co!).

### DIFF
--- a/tracer.js
+++ b/tracer.js
@@ -25,29 +25,38 @@ module.exports = (serviceName) => {
       }),
     });
 
-    let exporter;
+    let exporters = [];
+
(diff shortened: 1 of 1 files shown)


davehorton:
could fix these linting errors
```
/home/runner/work/jambonz-feature-server/jambonz-feature-server/tracer.js
Error:   28:9   error  'exporters' is never reassigned. Use 'const' instead  prefer-const
Error:   48:23  error  Expected parentheses around arrow function argument   arrow-parens
✖ 2 problems (2 errors, 0 warnings)
  2 errors and 0 warnings potentially fixable with the `--fix` option.
Error: Process completed with exit code 1.
```

zscgeek:
@davehorton - Updated!

davehorton (approved):

PR #972, opened 2024-11-13
unbridge dial ep with caller ep to avoid media release when referHook

xquanluu:


davehorton (approved):

vdharashive:
@xquanluu this is working as expected can you please merge it

Issue #955, opened 2024-10-31
feat: Inband tone dtmf needs to be supported 


vdharashive:
Carrier option can have which dtmf type supported this will give flexibility to configure based on need

PR #976, opened 2024-11-15
support enable dtmf tone

xquanluu:
#955

### DIFF
--- a/lib/session/call-session.js
+++ b/lib/session/call-session.js
@@ -30,6 +30,7 @@ const {
 } = require('../config');
 const bent = require('bent');
 const BackgroundTaskManager = require('../utils/background-task-manager');
+const dbUtils = require('../utils/db-utils');
 const BADPRECONDITIONS = 'preconditions not met';
 const CALLER_CANCELLED_ERR_MSG = 'Response not sent due to unknown transaction';
 
@@ -2060,6 +2061,8 @@ Duration=${duration} `
         };
       }
 
+      this._enableInbandDtmf(this.ep);
+
       // we are going from an early media connection to answer
       if (this.direction === CallDirection.Inbound) {
         // only do this for inbound call.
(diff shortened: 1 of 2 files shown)


lib/session/call-session.js
```diff
@@ -2601,6 +2604,26 @@ Duration=${duration} `
...
+    const voip_carrier_sid = this.req.has('X-Voip-Carrier-Sid') ? this.req.get('X-Voip-Carrier-Sid') :
+      this.req.has('X-Requested-Carrier-Sid') ? this.req.get('X-Requested-Carrier-Sid') : null;
+    if (voip_carrier_sid) {
+      const {lookupVoipCarrierBySid} = dbUtils(this.logger, this.srf);
+      const voipCarrier = await lookupVoipCarrierBySid(voip_carrier_sid);
+      if (voipCarrier?.dtmf_type === 'tones') {
+        // https://developer.signalwire.com/freeswitch/FreeSWITCH-Explained/Modules/mod-dptools/6587132/#0-about
+        try {
+          ep.execute('start_dtmf');
+          ep.inbandDtmfEnabled = true;
```
davehorton, line 2619: why do we need to set this property on ep?

lib/session/call-session.js
```diff
@@ -2601,6 +2604,26 @@ Duration=${duration} `
     }
   }
 
+  async _enableInbandDtmf(ep) {
```
davehorton, line 2607: I think since we are going to call this on every call, but we rarely actually enable inband dtmf we should name it differently, maybe like "_enableInbandDtmfIfRequired"

lib/session/call-session.js
xquanluu: Because we are runinging this check on every new task as there is some usecase that the media endpoint might be already exist, might be not created yet, or being replaced.

so that If this flag is enabled for that endpoint, we don't need to set the start_dtmf anymore.

lib/session/call-session.js
xquanluu: fixed

vdharashive:
@xquanluu @davehorton this is working fine have tested this telecom sip trunk , this can be merged

davehorton (approved):

vdharashive:
@xquanluu can this be merged

PR #988, opened 2024-11-29
fixed iamrole from sessionToken to securityToken

xquanluu:


vdharashive:
@xquanluu @davehorton I have tested this and it's working fine

davehorton:
have we tested each of the different ways that we offer in the jambonz web portal to authenticate to aws ?  I want to make sure we did not break any of the other methods while fixing this one

xquanluu:
Waiting new release for speech utils to support aws instance profile

https://github.com/jambonz/speech-utils/pull/99

davehorton (approved):

PR #990, opened 2024-12-02
Fix/986

sammachin:


lib/utils/error.js
```diff
@@ -1,9 +1,16 @@
-class SpeechCredentialError extends Error {
+class NonFatalTaskError extends Error {
```
davehorton, line 1: I think I would prefer to have a subclass of NonFatalTaskError, e.g

```
class PlayUrlNotFound extends NonFatalTaskError {
  constructor(url) {
    super('File not found');
    this.url = url; 
  }
```

Then in play.js throw that `PlayUrlNotFound`.  That way in our log messages we would get the specific URL not found I think, instead of this

```
{"level":50,"time":1733165080947,"pid":65211,"hostname":"daveh-macbook-pro.local","callId":"1-1@172.38.0.10-31","callSid":"7135de45-436f-450a-a3ed-1d379e9a6e10","accountSid":"bb845d4b-83a9-4cde-a6e9-50f3743bab3f","callingNumber":"play_single_link","calledNumber":"16174000000","traceId":"00000000000000000000000000000000","err":{"type":"NonFatalTaskError","message":"File not Found","stack":"Error: File not Found\n    at TaskPlay.exec (/Users/dhorton/tmp/jambonz-feature-server/lib/tasks/play.js:75:13)\n    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)\n    at async InboundCallSession.exec (/Users/dhorton/tmp/jambonz-feature-server/lib/session/call-session.js:1084:11)"},"msg":"Error executing task  #0:1: play"}
```

davehorton: in CallSession we can still test `instanceOf NonFatalTaskError` though because that will return true when testing if B is an instanceOf A when B subclasses A

davehorton (changes requested):

davehorton (approved):

lib/tasks/play.js
```diff
@@ -66,8 +67,20 @@ class TaskPlay extends Task {
         }
       }
     } catch (err) {
-      if (timeout) clearTimeout(timeout);
       this.logger.info(err, `TaskPlay:exec - error playing ${this.url}`);
+      this.playComplete = true;
+      if (err.message === 'File not found') {
```
rammohan-y, line 72: The string 'File not found' should be 'File Not Found' as returned from drachtio-msrf change https://github.com/drachtio/drachtio-fsmrf/blob/v3.0.46/lib/endpoint.js#L345

Without this, the fix is not working

lib/utils/action-hook-delay.js
```diff
@@ -126,7 +126,12 @@ class ActionHookDelayProcessor extends Emitter {
     try {
       this._taskInProgress = makeTask(this.logger, t[0]);
       this._taskInProgress.disableTracing = true;
-      this._taskInProgress.exec(this.cs, {ep: this.ep});
+      this._taskInProgress.exec(this.cs, {ep: this.ep}).catch((err) => {
+        this.logger.info(err, 'ActionHookDelayProcessor#_onNoResponseTimer: error playing file');
```
rammohan-y, line 130: logging err object causing the entire trace in the log file, can we log err.message>

lib/tasks/play.js
```diff
@@ -66,8 +67,20 @@ class TaskPlay extends Task {
         }
       }
     } catch (err) {
-      if (timeout) clearTimeout(timeout);
       this.logger.info(err, `TaskPlay:exec - error playing ${this.url}`);
```
rammohan-y, line 70: instead of logging the entire err object, can we log the err.message?

lib/tasks/play.js
sammachin: fixed in https://github.com/jambonz/jambonz-feature-server/pull/990/commits/d30f00f369763a554d0d0ba50d342024231da0f1

lib/tasks/play.js
davehorton: this would be a good idea, stack traces are not helpful in this case and can confuse someone reading the logs @sammachin

lib/tasks/play.js
```diff
@@ -67,7 +67,7 @@ class TaskPlay extends Task {
         }
       }
     } catch (err) {
-      this.logger.info(err, `TaskPlay:exec - error playing ${this.url}`);
+      this.logger.info(err.message, `TaskPlay:exec - error playing ${this.url}`);
```
davehorton, line 70: please instead make it
```
this.logger.info(`TaskPlay:exec - error playing ${this.url}: ${err.message`);
```

lib/utils/action-hook-delay.js
```diff
@@ -127,7 +127,7 @@ class ActionHookDelayProcessor extends Emitter {
       this._taskInProgress = makeTask(this.logger, t[0]);
       this._taskInProgress.disableTracing = true;
       this._taskInProgress.exec(this.cs, {ep: this.ep}).catch((err) => {
-        this.logger.info(err, 'ActionHookDelayProcessor#_onNoResponseTimer: error playing file');
+        this.logger.info(err.message, 'ActionHookDelayProcessor#_onNoResponseTimer: error playing file');
```
davehorton, line 130: ditto the change above.  Whenever we put a variable in the first part of a log statement please surround it with braces, otherwise the properties merge into the other log message properties and are had to distinguish, ie the above would be
```
this.logger.info({msg: err.message}, 'ActionHookDelayProcessor#_onNoResponseTimer: error playing file');
```
but in this case I would prefer simply
```
this.logger.info(`ActionHookDelayProcessor#_onNoResponseTimer: error playing file: ${err.message}`);
```

davehorton (changes requested):

davehorton (approved):

PR #993, opened 2024-12-04
fix ConfirmCallSession cannot be played

xquanluu:


lib/session/call-session.js
```diff
@@ -1043,8 +1043,9 @@ class CallSession extends Emitter {
     this.logger.info({tasks: listTaskNames(this.tasks)}, `CallSession:exec starting ${this.tasks.length} tasks`);
 
     // calculate if inbandDTMF tone is used
-    const voip_carrier_sid = this.req.has('X-Voip-Carrier-Sid') ? this.req.get('X-Voip-Carrier-Sid') :
-      this.req.has('X-Requested-Carrier-Sid') ? this.req.get('X-Requested-Carrier-Sid') : null;
+    // For ConfirmCallSession, req is not available and we don't need to enable DTMF tone for ConfirmCallSession
```
davehorton, line 1046: what is the issue with DTMF?  It seems like we are removing a feature from confirm session calls, why is that needed?

xquanluu: when new ConfirmCallSesson object, feature server does not assign req to the ConfirmCallSesson object, means inbandDTMF have not enabled for ConfirmCallSesson from the begining.

normally, feature server will do:

```
const session = new InboundCallSession(params);
session.req = req;
```

but this is not happening for ConfirmCallSession

xquanluu: and ConfirmCallSession cannot use gather verb

davehorton: why can't ConfirmCallSession use a gather?  I think that is as prime use case of a confirm call

xquanluu: I think I'm wrong there.
Fixed review comments, assign request to ConfirmCallSesson when new object for that.

vdharashive:
@xquanluu @davehorton this has been fixed ,it can be merged

Issue #996, opened 2024-12-10
Issue: Feature Server restarts on sip-refer notify webhook error


sammachin:
I've identified a scenario where the customers application crashes mid call and then the feature-server sends a webhook for a sip_refer. This results in UND_ERR_CLOSED from undici as it see the HTTP connection to the application as closed but this error when surfaced up to jambonz is not handled, thus the application quits.

Log entries showing the error and the restart
```
{"timestamp":"2024-12-09T09:43:09.317Z","errorGroups":[{"id":"CPiaqfXUoKuAOA"}],"log":{"msg":"web callback returned unexpected error","callId":"ca745f95-30b4-123e-90b8-42010a9c0004","url":"/jambonz-status","callSid":"b424f454-2ca9-4855-a804-d91a775e531c","level":50,"calledNumber":"9000","accountSid":"6fb344fd-f3b9-463a-b3bb-3ba8561590cd","baseUrl":"http://bot-router.apps",
	"err":
	{"message":"The client is closed",
		"type":"ClientClosedError",
		"code":"UND_ERR_CLOSED",
		"stack":"
			ClientClosedError: The client is closed\n    
				at Client.dispatch (/opt/app/node_modules/undici/lib/dispatcher/dispatcher-base.js:174:15)\n    
				at Client.request (/opt/app/node_modules/undici/lib/api/api-request.js:203:10)\n    
				at /opt/app/node_modules/undici/lib/api/api-request.js:196:15\n    
				at new Promise (<anonymous>)\n    
				at Client.request (/opt/app/node_modules/undici/lib/api/api-request.js:195:12)\n    
				at HttpRequestor.request (/opt/app/lib/utils/http-requestor.js:183:24)\n    
				at TaskSipRefer.performAction (/opt/app/lib/tasks/task.js:168:46)\n    
				at TaskSipRefer._handleNotify (/opt/app/lib/tasks/sip_refer.js:97:22)\n    
				at process.processTicksAndRejections 
				(node:internal/process/task_queues:95:5)",
		"name":"ClientClosedError"},
		"pid":1,
		"callingNumber":"[REDACTED]",
		"traceId":"12066ae4d6ef626b4339250f6a69c3ff",
		"hostname":"feature-server-6f8c7d57f9-zcbv8"
}}

	node:internal/process/promises:288
		triggerUncaughtException(err, true /* fromPromise */);
[...]

PR #1002, opened 2024-12-11
catch error

sammachin:


sammachin:
fixes #996

lib/tasks/sip_refer.js
```diff
@@ -94,7 +94,11 @@ class TaskSipRefer extends Task {
         }
         if (status >= 200) {
           this.referSpan.setAttributes({'refer.finalNotify': status});
-          await this.performAction({refer_status: 202, final_referred_call_status: status});
+          await this.performAction({refer_status: 202, final_referred_call_status: status})
+            .catch((err) => {
+              this.logger.error(err, 'TaskSipRefer:exec - error performing action finalNotify');
+              this.notifyTaskDone();
```
davehorton, line 100: It seems that `this.notifyTaskDone()` will be called twice in the case of an error, because it would have already been called from the line below.  So I think this can be removed

davehorton (changes requested):

lib/tasks/sip_refer.js
sammachin: ah yes we're not throwing a new error so it will continue through,

davehorton (approved):

PR #1004, opened 2024-12-12
fix say verb should fallback to 2nd vendor if response_code is not 2xx

xquanluu:
[09:18:46.199] ^[[34mDEBUG^[[39m (1001): ^[[36mSay got playback-stop^[[39m
    callId: "28be0b9d-32d2-123e-50bf-064156f7e4f7"
    callSid: "360f1cd2-9e5b-4cf5-865e-706c00bea293"
    evt: {
      "variable_tts_engine": "whisper",
      "variable_tts_voice": "alloy",
      "variable_tts_whisper_response_code": "0"
    }
    
    when we get playback-stop event with variable_tts_whisper_response_code = 0, say verb is not fallback to 2nd vendor

### DIFF
--- a/lib/tasks/say.js
+++ b/lib/tasks/say.js
@@ -193,8 +193,25 @@ class TaskSay extends TtsTask {
               }).catch((err) => this.logger.info({err}, 'Error adding file to cache'));
             }
 
+            function extractResponseCode(evt) {
+              for (const key in evt) {
+                if (key.startsWith('variable_tts_') && key.endsWith('_response_code')) {
+                  return evt[key];
+                }
+              }
+              return 0;
+            }
+
+            const response_code = extractResponseCode(evt);
+
             if (this._playResolve) {
-              evt.variable_tts_error ? this._playReject(new Error(evt.variable_tts_error)) : this._playResolve();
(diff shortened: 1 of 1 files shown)


lib/tasks/say.js
```diff
@@ -193,8 +193,25 @@ class TaskSay extends TtsTask {
...
+            function extractResponseCode(evt) {
+              for (const key in evt) {
+                if (key.startsWith('variable_tts_') && key.endsWith('_response_code')) {
+                  return evt[key];
+                }
+              }
+              return 0;
+            }
+
+            const response_code = extractResponseCode(evt);
```
davehorton, line 205: will this work for all vendors?  What about vendors, now or future that don't use http?

xquanluu: This will work with all vendors that use http and return variable_tts_<vendor> _response_code.
to support all case, we should keep the legacy code and do trust on variable_tts_error only

inside freeswitch modules, we have to update all vendors that if response code is not 2xx, variable_tts_error should have value so that fallback will happen on feature server.

davehorton:
why closed?

Issue #1034, opened 2025-01-13
Send error code when closing the websocket due to no response from far end


rammohan-y:
Send an error code 1001 (GOING_AWAY) to websocket close event when there is no response from the application with-in RESPONSE_TIMEOUT_MS.

PR #1035, opened 2025-01-13
feat/1034: sending socket close code when there is no response from the websocket

rammohan-y:
https://github.com/jambonz/jambonz-feature-server/issues/1034

Sending error code to socker close event when there is no response from the websocket app

### DIFF
--- a/lib/middleware.js
+++ b/lib/middleware.js
@@ -1,5 +1,5 @@
 const uuidv4 = require('uuid-random');
-const {CallDirection, AllowedSipRecVerbs} = require('./utils/constants');
+const {CallDirection, AllowedSipRecVerbs, WS_CLOSE_CODES} = require('./utils/constants');
 const {parseSiprecPayload} = require('./utils/siprec-utils');
 const CallInfo = require('./session/call-info');
 const HttpRequestor = require('./utils/http-requestor');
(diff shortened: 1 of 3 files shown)


rammohan-y:
@davehorton please review this PR

cc @vdharashive

lib/middleware.js
```diff
@@ -460,7 +460,7 @@ module.exports = function(srf, logger) {
       }).catch((err) => this.logger.info({err}, 'Error generating alert for parsing application'));
       logger.info({err}, `Error retrieving or parsing application: ${err?.message}`);
       res.send(480, {headers: {'X-Reason': err?.message || 'unknown'}});
-      app.requestor.close();
+      app.requestor.close(WS_CLOSE_CODES.GoingAway);
```
davehorton, line 463: @rammohan-y In this case we received a payload that is not a valid jambonz application.  That seems like it would map more correctly to a code of 1003
```
1003 indicates that an endpoint is terminating the connection
      because it has received a type of data it cannot accept
```
https://www.rfc-editor.org/rfc/rfc6455.html#section-7.4.1

davehorton (changes requested):

davehorton (approved):

PR #1048, opened 2025-01-18
Member mute on conference

avoylenko:
It seems like mute/unmute was never working for **Conference** task, so fixing this.


davehorton (approved):

lib/tasks/conference.js
```diff
@@ -549,8 +549,8 @@ class Conference extends Task {
   /**
    * mute or unmute side of the call
    */
-  mute(callSid, doMute) {
-    this.doConferenceMute(this.callSession, {conf_mute_status: doMute});
+  async mute(callSid, doMute) {
```
davehorton, line 552: why async when you are not awaiting on a promise anywhere

avoylenko: It not true, we are attaching the catch block here: https://github.com/jambonz/jambonz-feature-server/blob/7fa0041f6bf7692a8eb352c799fe0c7a36db61d3/lib/session/call-session.js#L1531

davehorton:
please rebase

lib/tasks/conference.js
davehorton: I still don't agree on this actually.

in line 1531 we call `task.mute` expecting to receive a Promise.
however `mute()` currently doesnt return anything
it does call `doConferenceMute()` which returns nothing.

The fact that we expect a Promise in line 1531 does not mean that the called function (`task.mute()`) needs to be declared async -- it can return a Promise without being declared async.  It only needs to be declared async if it uses await.

I prefer not to declare a function aysnc unless it uses await syntax within the function body.  Let me know if you spot problems in my reasoning.

davehorton (changes requested):

lib/tasks/conference.js
avoylenko: @davehorton you are 100% right, we don't need the async function here, but removing it will force us to change the mute function for dial verb. See  https://github.com/jambonz/jambonz-feature-server/blob/8bed44cce3446dd2f632e9d6fe49c1c95fa70a37/lib/tasks/dial.js#L336

avoylenko: Extra info, the mute function is being called here based on active task https://github.com/jambonz/jambonz-feature-server/blob/8bed44cce3446dd2f632e9d6fe49c1c95fa70a37/lib/session/call-session.js#L1524

davehorton: I really dont see why that would force a change anywhere else.

avoylenko: The `_lccMuteStatus` function calls the `mute` function based on session active task and since the function inside the dial task returns a promise(we have an await function there), we have to return a promise for the conference task as well.

avoylenko: @davehorton do you need anything else or PR can be accepted?

davehorton (approved):

avoylenko:
@davehorton could you please merge this?

davehorton (approved):

Issue #1059, opened 2025-01-28
Improve TTS caching granularity


javibookline:
While doing some testing, we noticed that changing some TTS options in the webapp portal did not generate a new audio but picked up old cached prompts generated with other TTS options.

For example, changing the elevenlabs model for another one, which should generate a very different sounding voice, hit the same cache file as when the audio was generated using another model.

https://github.com/jambonz/jambonz-feature-server/blob/16e044cabfecf8800a62c73c47f289810551979a/lib/tasks/say.js#L237

It seems that the evt.variable_tts_cache_filename does not take this differences into account. Maybe using a hash of the full request to the TTS service would allow for better granularity of caching and avoid these false cache hits.

xquanluu:
closed as PR is merged

PR #1062, opened 2025-02-02
support caching tts audio with model/model_id

xquanluu:
#1059


xquanluu:
Waiting for speech-utils to be released

Issue #997, opened 2024-12-10
Dial with transcribe at end of call transcript are not generated for last dialog 


vdharashive:
there is communication between customer and agent using call with bridge using dial verb with transcribe 

in this case the conversation between both party is transcribe just at end of the call 

Root cause is, as soon as the call got disconnected, jambonz is terminating the connection with ASR engines.  But ASR engine is taking some x seconds of time to give the transcription of the last spoken words.  Since jambonz disconnects the call, those words are not reaching to jambonz, hence those words are missing.

PR #1073, opened 2025-02-11
fixed dial transcribe is not able to receive final transcribe when closing the call

xquanluu:
#997


lib/config.js
```diff
@@ -139,6 +139,9 @@ const JAMBONES_USE_FREESWITCH_TIMER_FD = process.env.JAMBONES_USE_FREESWITCH_TIM
 const JAMBONES_DIAL_SBC_FOR_REGISTERED_USER = process.env.JAMBONES_DIAL_SBC_FOR_REGISTERED_USER || false;
 const JAMBONES_MEDIA_TIMEOUT_MS = process.env.JAMBONES_MEDIA_TIMEOUT_MS || 0;
 const JAMBONES_MEDIA_HOLD_TIMEOUT_MS = process.env.JAMBONES_MEDIA_HOLD_TIMEOUT_MS || 0;
+// jambonz
+const JAMBONES_DIAL_TRANSCRIBE_EP_DESTROY_DELAY_MS =
+  process.env.JAMBONES_DIAL_TRANSCRIBE_EP_DESTROY_DELAY_MS || 1000;
```
davehorton, line 144: I would rather not have a default, since this changes existing behavior

lib/utils/place-outdial.js
```diff
@@ -358,6 +359,24 @@ class SingleDialer extends Emitter {
         this.logger.info(err, 'place-outdial:_configMsEndpoint - error enable inband DTMF');
       }
     }
+
+    const origDestroy = this.ep.destroy.bind(this.ep);
+    this.ep.destroy = async() => {
+      try {
+        if (this.dialTask.transcribeTask) {
```
davehorton, line 366: only do this if we have a nested transcribe **and** JAMBONES_DIAL_TRANSCRIBE_EP_DESTROY_DELAY_MS has been set

davehorton (changes requested):

lib/config.js
xquanluu: fixed

lib/utils/place-outdial.js
xquanluu: fixed

davehorton (approved):

davehorton:
rebase

vdharashive:
@xquanluu this can be merged it's working as expected

vdharashive:
@xquanluu can this be also made working for transcribe rather then linking it with dial verb

xquanluu:
Hi @vdharashive 

Now transcription is working with both caller/callee hangup the call. tested and it's working on jambonz.me

lib/session/call-session.js
```diff
@@ -2838,6 +2839,24 @@ Duration=${duration} `
...
+    this.ep.destroy = async() => {
+      try {
+        if (this.currentTask.name === TaskName.Transcribe && JAMBONES_TRANSCRIBE_EP_DESTROY_DELAY_MS) {
+          // transcribe task is being used, wait for some time before destroy
+          // if final transcription is received but endpoint is already closed,
+          // freeswitch module will not be able to send the transcription
+
+          this.logger.info('callSession:_configMsEndpoint -' +
+            ' transcribe task, wait for some time before destroy');
+          await new Promise((resolve) => setTimeout(resolve, JAMBONES_TRANSCRIBE_EP_DESTROY_DELAY_MS));
```
davehorton, line 2853: I prefer to just add a `sleepFor` function at the top of the file and use that, it's more understandable
```
const sleepFor = (ms) => new Promise((resolve) => setTimeout(() => resolve(), ms));
```

Now I see that in dequeue.js we already did something similar.  Please move the function to utils so we can just require it anywhere needed

davehorton (changes requested):

lib/session/call-session.js
xquanluu: fixed.

lib/utils/jambonz.js
```diff
@@ -0,0 +1,5 @@
+
```
davehorton, line 1: can we instead name this file `/utils/helpers.js`

davehorton (changes requested):

lib/utils/jambonz.js
xquanluu: fixed

davehorton (approved):

PR #1075, opened 2025-02-11
support create call with target.proxy

xquanluu:


xquanluu:
It's working with this payload:

```
"to": {
          "type": "sip",
          "sipUri": "sip:1234123@50.51.52.53",
          "proxy": "sip:51.52.53.54"
      },
```

davehorton (approved):

Issue #1094, opened 2025-02-24
Send jambonz:error websocket response when payload is not parsable


rammohan-y:
When an invalid command is sent to webscoket requester, send a jambonz:error back if the payload is not parsable, currently the invalid payload is only logged without sending any jambonz:error message back to the client

vdharashive:
@davehorton @xquanluu can you please review the fix

PR #1095, opened 2025-02-24
sending jambonz:error when the incoming message is not parsable

rammohan-y:
sending jambonz:error when the incoming message is not parsable
https://github.com/jambonz/jambonz-feature-server/issues/1094

### DIFF
--- a/lib/utils/ws-requestor.js
+++ b/lib/utils/ws-requestor.js
@@ -433,7 +433,8 @@ class WsRequestor extends BaseRequestor {
       this.logger.info({err, content}, 'WsRequestor:_onMessage - invalid incoming message');
       const params = {
         msg: 'InvalidMessage',
-        details: err.message
+        details: err.message,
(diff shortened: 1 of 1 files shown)


lib/utils/ws-requestor.js
```diff
@@ -431,6 +431,12 @@ class WsRequestor extends BaseRequestor {
       }
     } catch (err) {
       this.logger.info({err, content}, 'WsRequestor:_onMessage - invalid incoming message');
+      const params = {
+        msg: 'InvalidMessage',
+        details: err.message
```
davehorton, line 436: I suggest including 'content' as well in the error so the sender can optionally see the source message that caused the error

davehorton (changes requested):

lib/utils/ws-requestor.js
rammohan-y: @davehorton added content to jambonz:error payload and also written an alert without content, hope this is okay

PR #1096, opened 2025-02-25
if redirect to new server update requestor for baseURL

### LINKED ISSUE #1079: redirect doesn't update baseURL for future verbs
Say you have an application configured with a callHook at https://hostone.example.com/call 
and that callHook returns a `redirect` verb to a new host & path  say https://hosttwo.example.com/call

If the callHook on hosttwo then uses a verb like gather and sets its action hook as relative eg `/input` the input from the gather is sent to https://hostone.example.com/input

I would expect it to have gone to hosttwo

It appears that redirect does not update the baseurl for the call session which is what I would expect

sammachin:
fixes #1079 

This doesn't update the notifier object so the call status webhooks will still go to the original app url, but where the callHook is updated any future relative URLs in the new app will goto that server.

### DIFF
--- a/lib/tasks/redirect.js
+++ b/lib/tasks/redirect.js
@@ -1,6 +1,8 @@
 const Task = require('./task');
 const {TaskName} = require('../utils/constants');
 const WsRequestor = require('../utils/ws-requestor');
+const URL = require('url');
+const HttpRequestor = require('../utils/http-requestor');
 
 /**
  * Redirects to a new application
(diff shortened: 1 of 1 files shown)


lib/tasks/redirect.js
```diff
@@ -24,6 +26,21 @@ class TaskRedirect extends Task {
       } catch (err) {
         this.logger.info(err, `Task:performAction error redirecting to ${this.actionHook}`);
       }
+    } else if (cs.application.requestor._isAbsoluteUrl(this.actionHook)) {
+      const baseUrl =  this.cs.application.requestor.baseUrl;
+      const newUrl = URL.parse(this.actionHook);
+      const newBaseUrl = newUrl.protocol + '//' + newUrl.host;
+      if (baseUrl != newBaseUrl) {
```
davehorton, line 33: just to be sure, you have tested where an absolute url is provided but has the same base url as the current, and seen that it does not update the base url?

sammachin: yes

sammachin:
I've also PR'd the docs with this change but I guess we want to wait till its deployed to merge that https://github.com/jambonz/jambonz-fern-config/pull/15

sammachin:
done

davehorton (approved):

Issue #1103, opened 2025-03-03
API Hangup sends wrong `call_terminated_by` in statusHook


sammachin:
When terminating a call using the API eg live call control, PUT callstatus completed, the call status webhook has 'caller' in the call_terminated_by field this should be 'jambonz'

To reprodice
Create a call and then terminate it by sending a PUT

sammachin:
fixed in #1104

PR #1104, opened 2025-03-03
call jambonzHangup when API ends call

sammachin:
fixes 1103


davehorton:
Note: in commit messages if you use the hash sign, eg 

fixes #1103 

instead of fixes 1103
It will create a link to the issue

sammachin:
sorry yes that was a typo

> Note: in commit messages if you use the hash sign, eg
> 
> fixes #1103
> 
> instead of fixes 1103 It will create a link to the issue

davehorton (approved):

PR #1111, opened 2025-03-06
Feature: add digitCount to amd-utils

sammachin:
Adds a new feature to AMD,

Many voicemail systems (especially on cellular) include the called number as part of their default greeting, the transcription engines return these digits as a single "word" eg `07790123900 is not availble` rather than spacing out each digit as a new word, therefore the `thresholdWordCount` is often not reached for what is in effect a long message.

This adds a new feature that allows you to set a `digitCount` property on the AMD check and if there is a string of digits detected it will trigger on that.
By default it is off (digitCount = 0) so won't trigger. 
The length is adjustable for different scenarios and counties, with UK numbers I've found that 5-6 is a good value.

There are _some_ older people in the UK at least that still answer their landlines with the phone number eg "London 1234" so like all AMD this won't be rock solid but I think its a useful addition.

There's a corresponding PR in verb-specifications too (https://github.com/jambonz/verb-specifications/pull/88) and if we decide to go with it, and I'll PR the docs when merged

### DIFF
--- a/lib/utils/amd-utils.js
+++ b/lib/utils/amd-utils.js
@@ -45,6 +45,7 @@ if (VMD_HINTS_FILE) {
   });
 }
 
+
 class Amd extends Emitter {
   constructor(logger, cs, opts) {
     super();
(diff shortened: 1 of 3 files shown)


davehorton:
looks good to me, I will merge verb-specs when out of draft status

davehorton:
rebase and update to verb-specs and should be good.  Nice feature!

sammachin:
done

davehorton (approved):

sammachin:
It aint done till its documented ;) https://github.com/jambonz/jambonz-fern-config/pull/21

Issue #1121, opened 2025-03-17
feature server in cluster mode cannot isolate bad instance


xquanluu:
Feature server is running on a machine with:
1/ 4 feature server node js app
2/ 1 drachtio server
3/ 1 freeswitch

From start of the machine there are only 2 feature servesr are able to connect with freeswitch and the orther 2 are not able to connect to freeswitch.

in the redis, the drachtio server IP will be available for handling incoming call.

When drachtio server receive INVITE request, drachtio server will notify feature server, some call will be handle by good feature server, other will be rejected because of bad feature server.

feature server should reconnect to freeswitch in this case:

Also There is another issue is fixed on this PR: https://github.com/drachtio/drachtio-fsmrf/pull/107

xquanluu:
closing this as PR is merged

PR #1123, opened 2025-03-18
only connect to drachtio server if connected to freeswitch

xquanluu:
#1121


xquanluu:
tested wih restart both freeswitch and drachtio server, feature server are able to issolate himself out of cluster and record back.

davehorton (approved):

PR #1124, opened 2025-03-18
Feat 1120 video call - remove video media from SDP if the call is audio call

rammohan-y:
check if the call is not a video call, if yes, then remove video media from the sdp, when we enable VP8 in codecs in freeswitch, even webrtc audio call is sending video media in the SDP


rammohan-y:
@davehorton introduced JAMBONES_VIDEO_CALLS_ENABLED_IN_FS environment variable, if this variable is set to true then we will call removeVideoSdp

davehorton (approved):

rammohan-y:
@davehorton fixed jslint errors after your approval, sorry for that

davehorton (approved):

Issue #1106, opened 2025-03-04
deepgramOptions.model not being used


sammachin:
In a `gather` verb using deepgram for the recognizer and setting a `deepgramOptions.model` property of `nova-3` The model is not being used, instead the default model is used by calling 
https://github.com/jambonz/jambonz-feature-server/blob/a5e778d7f3e748712aeb3cf1abe246feaaaa2134/lib/utils/transcription-utils.js#L794

Adding logging at L786 shows that `model` is undefined at this point despite being in the verb.

```
10|jambonz-feature-server  | DEEPGRAM MODEL: undefined
10|jambonz-feature-server  | {"level":30,"time":1741082439427,"pid":264662,"hostname":"jb","callId":"5cd5fa9d-7382-123e-6991-960003fc59c6","callSid":"ea1c47e5-3c70-4877-b48b-6e08eabb7dcb","accountSid":"72c5c38f-9bba-40ce-aa83-aaa6be55e1b5","callingNumber":"+441934834213","calledNumber":"447973994474","traceId":"ddc4e3e58c0fd15f6484a06a9e1c8eed","msg":"CallSession:exec starting task #0:4: gather"}
10|jambonz-feature-server  | {"level":30,"time":1741082439429,"pid":264662,"hostname":"jb","callId":"5cd5fa9d-7382-123e-6991-960003fc59c6","callSid":"ea1c47e5-3c70-4877-b48b-6e08eabb7dcb","accountSid":"72c5c38f-9bba-40ce-aa83-aaa6be55e1b5","callingNumber":"+441934834213","calledNumber":"447973994474","traceId":"ddc4e3e58c0fd15f6484a06a9e1c8eed","msg":"Curreet BaseURL is https://jambonz.sammachin.com"}
[...]

PR #1126, opened 2025-03-24
use deepgramOptions.model

sammachin:
This uses the value of deepgramOptions.model in the recognizer as per the docs,
It will still fallback to using recognizer.model if that is set but the deepgramOptions will take priortiy if both are set as that feels like a more specific setting but doesn't break any existing apps.

### DIFF
--- a/lib/utils/transcription-utils.js
+++ b/lib/utils/transcription-utils.js
@@ -789,7 +789,7 @@ module.exports = (logger) => {
       };
     }
     else if ('deepgram' === vendor) {
-      let {model} = rOpts;
+      let model = rOpts.deepgramOptions.model || rOpts.model
       const {deepgramOptions = {}} = rOpts;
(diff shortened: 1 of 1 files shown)


sammachin:
fixes #1106

lib/utils/transcription-utils.js
```diff
@@ -789,7 +789,7 @@ module.exports = (logger) => {
       };
     }
     else if ('deepgram' === vendor) {
-      let {model} = rOpts;
+      let model = rOpts.deepgramOptions.model || rOpts.model;
```
davehorton, line 792: I think needs to be `rOpts.deepgramOptions?.model` in case the user does not provide a deepgramOptions property?

davehorton (changes requested):

lib/utils/transcription-utils.js
sammachin: yes good catch, i didn't think about that scenario, tried other comibantion of model

davehorton (approved):

Issue #1135, opened 2025-04-03
Allowing hangup for siprec call


rammohan-y:
Can we allow "hangup" verb in a siprec call? This can be useful if we want to terminate a siprec call after it is initiated.

PR #1136, opened 2025-04-03
Allow hangup verb on siprec call

rammohan-y:
https://github.com/jambonz/jambonz-feature-server/issues/1135

### DIFF
--- a/lib/utils/constants.json
+++ b/lib/utils/constants.json
@@ -32,7 +32,7 @@
     "Tag": "tag",
     "Transcribe": "transcribe"
   },
-  "AllowedSipRecVerbs": ["answer", "config", "gather", "transcribe", "listen", "tag"],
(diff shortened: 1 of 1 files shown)


lib/utils/constants.json
```diff
@@ -32,7 +32,7 @@
     "Tag": "tag",
     "Transcribe": "transcribe"
   },
-  "AllowedSipRecVerbs": ["answer", "config", "gather", "transcribe", "listen", "tag"],
+  "AllowedSipRecVerbs": ["answer", "config", "gather", "transcribe", "listen", "tag", "hangup"],
```
davehorton, line 35: I suppse we we should allow "sip:decline" as well

davehorton (changes requested):

lib/utils/constants.json
rammohan-y: @davehorton added sip:decline to the array

davehorton (approved):

PR #1144, opened 2025-04-09
fix initial message format in elevenlabs

sammachin:
Fixes an issue with the inital message that we're sending to elvenlabs, the current message format is incorrect as per their docs: https://elevenlabs.io/docs/conversational-ai/api-reference/conversational-ai/websocket#send.Conversation%20Initiation%20Client%20Data

We are sending:
```
{
  "type": "conversation_initiation_client_data",
  "conversation_initiation_client_data": {
    "dynamic_variables": {
      "user": "Sam"
    }
  }
}
```
And it should be 
```
{
  "type": "conversation_initiation_client_data",
  "dynamic_variables": {
    "user": "Sam"
  }
}
```

Elevenlabs still don't seem to be recognising the dynamic variables or conversation_config_override though so may need further fixes based on their feedback.


davehorton:
@sammachin can you rebase?  Did this pass your testing?

davehorton (approved):

PR #1148, opened 2025-04-11
return default empty conversation_initiation_client_data message

sammachin:
This will send an initiation message with just the type field if no options have been specified by the user which will then allow the freeswitch module to start sending audio.
You can see that the message is sent in the freeswitch log:
`2025-04-11 13:28:02.427817 89.73% [DEBUG] elevenlabs_glue.cpp:369 elevenlabs_s2s_send_client_event: {"type":"conversation_initiation_client_data"}`

Tested with both conversation_initiation_client_data data and with just the sampleRate in llmOptions


davehorton:
Closing in favor of https://github.com/jambonz/jambonz-feature-server/pull/1149

PR #1175, opened 2025-04-28
send env_vars in callHook

sammachin:


lib/middleware.js
```diff
@@ -417,10 +417,18 @@ module.exports = function(srf, logger) {
             ...(app.fallback_speech_recognizer_language && {fallback_language: app.fallback_speech_recognizer_language})
           }
         };
+        let env_vars;
+        if (app.env_vars) {
+          const d_env_vars = JSON.parse(decrypt(app.env_vars));
```
davehorton, line 422: I think we better wrap this in a try-catch

davehorton (changes requested):

davehorton:
rebase needed

lib/middleware.js
sammachin: fixed in https://github.com/jambonz/jambonz-feature-server/pull/1175/commits/3b4cfd5351308505a44ba86db142e222d1609a39

Issue #1179, opened 2025-04-30
play verb kill is not working sporadically


rammohan-y:
we are facing issues will play kill when a audio file of 90+ seconds is played, after issuing play command with queueCommand as true, subsequent commands which are queueCommand as false are not killing the audio.

the same code is attached in https://jambonz.freshdesk.com/support/tickets/846

```
const http = require('http')
const express = require('express')
const port = process.env.WS_PORT || 3011;
const port2 = process.env.WS_PORT2 || 3021;
const bodyParser = require('body-parser')

const expressApp = express();
const server = http.createServer(expressApp)
expressApp.use(bodyParser.json())


const app = express();
app.use(bodyParser.json())

app.listen(port2,
    function (err) {
        if (err) console.log(err);
        console.log("Server listening on PORT", port2);
    });
expressApp.post('/register',
    (req, res) => {
        console.log("inside register")
        let msg = {
            'status': 'ok',
        };
        res.end(JSON.stringify(msg));
    });
expressApp.get('/register',
    (req, res) => {
        let msg = {
            'status': 'ok',
        };
        res.end(JSON.stringify(msg));
    });
server.listen(port, function () {
    console.log(`**** SAVG Listener Server running on port ${port} **** `)
})


const WebSocket = require('ws')
const wss = new WebSocket.Server({ server, path: "/socket" });
const { WebhookResponse } = require('@jambonz/node-client');

function ack(ws, msgid, res) {
    if (res) {
        res = res instanceof WebhookResponse ? res.toJSON() : res;
    }
    let msg = {
        type: 'ack',
        msgid
    };
    if (res) msg = { ...msg, data: res };
    console.log("sending ack message ", JSON.stringify(msg))
    ws.send(JSON.stringify(msg));
}
function sendCommand(ws, payload, queueCommand = false) {
    payload = payload instanceof WebhookResponse ? payload.toJSON() : payload;

    const msg = {
        type: 'command',
        command: 'redirect',
        queueCommand,
        data: payload
    };
    ws.send(JSON.stringify(msg));
}
wss.on('connection', function (ws) {
    ws.on('close', function (data) {
        console.log('Client disconnected', data);
        // Add any cleanup code here if needed
    });
    ws.on('message', async function (dataStr) {
        let wsData = JSON.parse(dataStr)
        console.log("wsData.type", wsData)
        const { type, msgid, hook, call_sid, data } = wsData;
        const { from, to } = data;
[...]

PR #1180, opened 2025-05-01
play verb: fix race condition where play canceled immediately after s…

davehorton:
…tart


davehorton:
fix for https://github.com/jambonz/jambonz-feature-server/issues/1179

rammohan-y:
The issue is still reproducing in our non-prod environment

PR #1183, opened 2025-05-06
Feat 1179 race issue with play verb

rammohan-y:
Feat 1179 race issue with play verb

### DIFF
--- a/lib/session/call-session.js
+++ b/lib/session/call-session.js
@@ -23,6 +23,7 @@ const HttpRequestor = require('../utils/http-requestor');
 const WsRequestor = require('../utils/ws-requestor');
 const ActionHookDelayProcessor = require('../utils/action-hook-delay');
 const TtsStreamingBuffer = require('../utils/tts-streaming-buffer');
+const StickyEventEmitter = require('../utils/sticky-event-emitter');
 const {parseUri} = require('drachtio-srf');
 const {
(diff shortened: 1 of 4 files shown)


davehorton:
My concern about this PR is the possibility of memory leaks.  The caches and stored listeners create possibility for references to persist and for memory not to be cleared.  Testing to be sure this is not the case is actually quite arduous.  At a minimum, there should be a 'destroy' method on the sticky emitter that releases clears the cache and releases all references to listeners.

Still, we are left with a fairly complex solution to the bug.  I am undecided whether this PR represents the solution or whether the [simpler PR here](https://github.com/jambonz/jambonz-feature-server/pull/1180) is less risky, while not quite as elegant.

rammohan-y:
I completely understand your observations, and will look into the clearing of cache with a destroy method.

Actually, regarding the complexity, it is not that complex as it looks like, we are making sure that a uuid_break is properly executed and ensures that the play verb causing the playing in freeswitch is indeed stopped

With the simple PR, I tested it and found that it behaved differently in different runtime scenarios, hence not sure what would be correct timeout value. The more is the timeout the more possibility of getting into other race conditions, the less the timeout value the more the possibility of not killing the audio

rammohan-y:
@davehorton please review the memory leak fix

Issue #1176, opened 2025-04-28
tts-stream not playing the stream when LLM is returning special characters in the stream


rammohan-y:
tts-stream not playing the stream when LLM is returning special characters in the stream, for example when you give input as "generate 10 words sentence enclosed in double quotes", the tokens generated by the LLM are not being played by tts-streaming

Example code below:

```
const OpenAI = require("openai");
const OPENAI_API_KEY = 'your-api-key; // replace with your API key
const openai = new OpenAI({ apiKey: OPENAI_API_KEY });
const service = ({logger, makeService}) => {
  const svc = makeService({path: '/llm-streaming'});
  
  svc.on('session:new', (session) => {
    
    console.log("inside llm-streaming");
    session.locals = {
      logger: logger.child({call_sid: session.call_sid})
    };
    logger.debug({session}, `new incoming call: ${session.call_sid}`);


    session
      .on('/speech-detected', onSpeechDetected.bind(null, session))
      .on('tts:streaming-event', onStreamingEvent.bind(null, session))
      .on('tts:user_interrupt', onUserInterrupt.bind(null, session))
      .on('close', onClose.bind(null, session))
      .on('error', onError.bind(null, session));

    try {
      session
        .config({
          synthesizer: {
            vendor: "deepgram",
            voice: "aura-asteria-en",
            language: "en-US"
          },
          ttsStream: {
            enable: true,
          },
          bargeIn: {
            enable: true,
            sticky: true,
            minBargeinWordCount: 1,
            actionHook: '/speech-detected',
            input: ['speech']
          }
        })
        //.say({text: 'Hi there, how can I help you today?'})
        .send();
    } catch (err) {
      session.locals.logger.info({err}, `Error to responding to incoming call: ${session.call_sid}`);
      session.close();
    }
  });
};

const onSpeechDetected = async(session, event) => {
  const {logger, client} = session.locals;
  const {speech} = event;

  session.reply();

  if (speech?.is_final) {
    const {transcript} = speech.alternatives[0];
    session.locals.user_interrupt = false;

    logger.info({transcript}, `session ${session.call_sid} received speech detected event`);

    const stream = await openai.chat.completions.create({
      model: "gpt-4.1",
      messages: [
        {
          role: "user",
         // content: transcript,
          // content: "ask me a question and place a single quote after the question mark",
        //  content: "ask me a question and place a double quotes after the question mark",
         content: "generate a double quoted string of 10 words",
        },
      ],
[...]

PR #1184, opened 2025-05-07
fixed tts streaming buffer cannot reset timeoutwhen lastUpdateTime is short

xquanluu:
#1176

### DIFF
--- a/lib/utils/tts-streaming-buffer.js
+++ b/lib/utils/tts-streaming-buffer.js
@@ -377,6 +377,7 @@ class TtsStreamingBuffer extends Emitter {
 
   _onTimeout() {
     this.logger.debug('TtsStreamingBuffer:_onTimeout Timeout waiting for sentence boundary');
+    this.timer = null;
     // Check if new text has been added since the timer was set.
     const now = Date.now();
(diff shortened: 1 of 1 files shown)


lib/utils/tts-streaming-buffer.js
```diff
@@ -377,14 +377,14 @@ class TtsStreamingBuffer extends Emitter {
 
   _onTimeout() {
     this.logger.debug('TtsStreamingBuffer:_onTimeout Timeout waiting for sentence boundary');
+    this.timer = null;
     // Check if new text has been added since the timer was set.
     const now = Date.now();
     if (now - this.lastUpdateTime < TIMEOUT_RETRY_MSECS) {
```
xquanluu, line 383: Hi @davehorton 
TIMEOUT_RETRY_MSECS = 3000
There is a case that the last token came shortly and then this timeout happen.
jambonz reset timer for another 3 seconds. and no more token come.

This make user wait for 3 seconds to hear back the audio. should we reduce this timeout to 1 second?

davehorton: agree

xquanluu: fixed

davehorton (approved):

PR #1188, opened 2025-05-09
Prioritize JAMBONES_LOGLEVEL over db setting

avoylenko:
The `JAMBONES_LOGLEVEL` env no longer has a weight and it was required just to init the logger. This PR implements comparing the `JAMBONES_LOGLEVEL` env with the setting in the database before deciding what log level should be system wide.  
Also, adding a message about changing the log level.

### DIFF
--- a/app.js
+++ b/app.js
@@ -30,6 +30,13 @@ const installSrfLocals = require('./lib/utils/install-srf-locals');
 const createHttpListener = require('./lib/utils/http-listener');
 const healthCheck = require('@jambonz/http-health-check');
 
+logger.on('level-change', (lvl, _val, prevLvl, _prevVal, instance) => {
+  if (logger !== instance) {
+    return;
+  }
+  logger.info('system log level %s was changed to %s', prevLvl, lvl);
+});
+
 // Install the srf locals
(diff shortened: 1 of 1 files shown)


davehorton:
can you provide more description on how this feature works?

avoylenko:
Prior this change the system log level was changing based on the `log_level` column from the `system_information` table and the log level could be just **info** or **debug**. The `JAMBONES_LOGLEVEL` env was ignored. 

This change adds the weight for the `JAMBONES_LOGLEVEL` env back, the highest log level between the `JAMBONES_LOGLEVEL` env and the `log_level` column will be applied. In other words, if the `JAMBONES_LOGLEVEL` env is set to **debug**, but the db setting is set to **info**, the resulting log level will be **debug**.

davehorton:
and specifically tell me more about this event, when it is fired and what the intent is
```
logger.on('level-change', ..
```

avoylenko:
@davehorton it's a listener function that triggers when the log level is changed.
https://github.com/pinojs/pino/blob/main/docs/api.md#level-change

davehorton (approved):

PR #1235, opened 2025-06-06
fix(create-call): fix url parsing

RohanSreerama5:
Closes https://jambonz.freshdesk.com/a/tickets/995

fix url parsing


davehorton:
@RohanSreerama5 were you able to test before and after this fix on the sample the customer provided?

Issue #1249, opened 2025-06-20
Add support for Ultravox preconfigured agents/tools


atemnorod01:
Please add support for Ultravox preconfigured agents to LLM verb.
https://docs.ultravox.ai/api-reference/agents/agents-calls-post

xquanluu:
Closing this ticket as PR is merged.

PR #1254, opened 2025-06-22
support ultravox agent id

xquanluu:
#1249


davehorton:
let me know when you have tested both the agent api and the backwards compatibility for calls api

xquanluu:
Hi @davehorton 

it's working and backwards compatibility for calls api

xquanluu:
jambonz application example code is here:
https://github.com/jambonz/ultravox-s2s-example/pull/8

davehorton (approved):

PR #1269, opened 2025-07-04
FD_1079: tts modules response_code = 0 should make say fail to exec.

xquanluu:


vdharashive:
@xquanluu this breaking basic say

xquanluu:
fixed it @vdharashive

PR #1283, opened 2025-07-13
Fix for issue where we should not delay from gather when JAMBONES_TRANSCRIBE_EP_DESTROY_DELAY_MS is set

rammohan-y:
We should not delay stopTranscription method when JAMBONES_TRANSCRIBE_EP_DESTROY_DELAY_MS is set

### DIFF
--- a/lib/tasks/gather.js
+++ b/lib/tasks/gather.js
@@ -385,6 +385,9 @@ class TaskGather extends SttTask {
       if (this.digitBuffer.length === 0 && this.needsStt) {
         // DTMF is higher priority than STT.
         this.removeCustomEventListeners();
+        // Fix for https://github.com/jambonz/jambonz-feature-server/issues/1281
+        // We should immediately call stop transcription from gather
+        // so that next gather can start transcription immediately
         ep.stopTranscription({
           vendor: this.vendor,
           bugname: this.bugname,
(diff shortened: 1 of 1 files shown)


vdharashive:
@xquanluu can you please review this pr,

This feature should be applicable for transcribe verb only

xquanluu:
Hi @rammohan-y 

could you please help to add comment code to explain why that change is required for only gather verb, that allow us later don't remove that code change.

lib/tasks/gather.js
```diff
@@ -1236,9 +1240,13 @@ class TaskGather extends SttTask {
...
     this.logger.info({evt}, `TaskGather:resolve with reason ${reason}`);
     if (this.needsStt && this.ep && this.ep.connected) {
+      // Fix for https://github.com/jambonz/jambonz-feature-server/issues/1281
+      // We should immediately call stop transcription from gather
+      // so that next gather can start transcription immediately
       this.ep.stopTranscription({
         vendor: this.vendor,
-        bugname: this.bugname
+        bugname: this.bugname,
+        gracefulShutdown: false
```
davehorton, line 1249: What is the purpose of adding gracefulShutdown to the api call here?  It is not used in Endpoint#stopTranscription

xquanluu: It is because:
1/ when JAMBONES_TRANSCRIBE_EP_DESTROY_DELAY_MS is enabled, feature server override stopTranscription function  from endpoint to delay sending stop transcribe command to freeswitch.
2/ In some case: gather verb, pause transcribe, fall over to new recognizer, stopTranscription should not do that kind of delay/gracefullShutdown action.
3/ in overrided stopTranscription , it uses that flag gracefulShuwdown  to do the delay stop command or not

davehorton (approved):

Issue #1285, opened 2025-07-15
TTS cache issue


rammohan-y:
In the recent fix for TTS cache issue, we have introduced an id in freeswitch module and propagating it to feature-server, so that in playback-stop we compare the id and if it doesn't match we will log a message "Say:exec discarding playback-stop for earlier play"

Couple of issues in the fix:

a. When TTS returns 404 or any other error, the playback-start is not invoked, but playback-stop is invoked with last generated id, this will fail the if condition at https://github.com/jambonz/jambonz-feature-server/blob/main/lib/tasks/say.js#L281 and unmatchedResponse will become true and the error is logged. We added an alert for this message, this message should be logged only when playbackId is there and does not match with the paybackId returned in the playback-stop event

b. As we have introduced a static variable (currentId) in freeswitch module, will it not have same value in a large volume concurrency scenario, which might cause same value to be returned in the playback-start and/or playback-stop events and can cause a raise condition triggering a possibilty of wrong file getting written to cache? If we introduce locking it might cause the entire system to become slow 

Instead of using an ID approach can we do below

c. Sending the text to freeswitch and returning the same text in playback-stop event and using this text value from the event in addFIleToCache method? - This seems to be very safe approach, as we are caching the audio content for the text that is used in freeswitch to generate the audio

@davehorton @xquanluu @vdharashive

PR #1286, opened 2025-07-15
fix for #1285

davehorton:


davehorton:
fixed in PR plus freeswitch-modules 2.3.5

Issue #1289, opened 2025-07-17
Whisper/Play fails


sammachin:
Trying to PUT a whisper to a call_sid with a play verb produces the below error, using a say with tts in the whisper works fine

```
8|jambonz-feature-server  | {"level":20,"time":1752744136863,"pid":317201,"hostname":"jb","body":{"whisper":{"verb":"play","url":"https://s3.sammachin.com/count30.wav"}},"msg":"got updateCall request"}
8|jambonz-feature-server  | TypeError: this.requestor is not a function
8|jambonz-feature-server  |     at InboundCallSession._lccWhisper (/home/admin/apps/jambonz-feature-server/lib/session/call-session.js:1651:31)
8|jambonz-feature-server  |     at InboundCallSession.updateCall (/home/admin/apps/jambonz-feature-server/lib/session/call-session.js:1883:21)
8|jambonz-feature-server  |     at /home/admin/apps/jambonz-feature-server/lib/http-routes/api/update-call.js:65:10
8|jambonz-feature-server  |     at Layer.handle [as handle_request] (/home/admin/apps/jambonz-feature-server/node_modules/express/lib/router/layer.js:95:5)
8|jambonz-feature-server  |     at next (/home/admin/apps/jambonz-feature-server/node_modules/express/lib/router/route.js:149:13)
8|jambonz-feature-server  |     at Route.dispatch (/home/admin/apps/jambonz-feature-server/node_modules/express/lib/router/route.js:119:3)
8|jambonz-feature-server  |     at Layer.handle [as handle_request] (/home/admin/apps/jambonz-feature-server/node_modules/express/lib/router/layer.js:95:5)
8|jambonz-feature-server  |     at /home/admin/apps/jambonz-feature-server/node_modules/express/lib/router/index.js:284:15
8|jambonz-feature-server  |     at param (/home/admin/apps/jambonz-feature-server/node_modules/express/lib/router/index.js:365:14)
8|jambonz-feature-server  |     at param (/home/admin/apps/jambonz-feature-server/node_modules/express/lib/router/index.js:376:14)
```

sammachin:
fixed in #1290

PR #1290, opened 2025-07-17
don't fetch if whisper is an object with a single verb in it

sammachin:
partial fix,

This allows a single object verb of play to be used in a whisper but fetching verbs from  a URL still fails


sammachin:
fixes #1289

davehorton (approved):

PR #1298, opened 2025-07-23
Update cron-jobs.js

### LINKED ISSUE #1294: Cache file in /tmp or /tmp/tts-cache-files/ are not delete with .r8 extension
Code written in this block handles only mp3 files 

https://github.com/jambonz/jambonz-feature-server/blob/bd1c763e7266aae0399e62a42e95a9c4397ea13b/lib/utils/cron-jobs.js#L51

Rather logic to be implemented start with tts-*.*

Also 

The filder location to be used as environment variable if its set JAMBONES_TMP_FOLDER

https://github.com/jambonz/speech-utils/blob/043642ea5fb42184dc512ca36a697af6d52212fe/lib/config.js#L12

Also in cron-jobs scheduled on os level need to have this change

sammachin:
fixes #1294
This relies on a new shell script which is added to the deployment tools repo.

Using a single script allows the cleanup logic to be the same regardless of if it is triggered from the feature server or from cron on the OS


davehorton:
I think the function needs to handle failure (eg when the script is not there) log an error, write an alarm, and not continue retrying at that point.

More generally, I dont know that I love creating a new dependency on an external script....we have no such dependencies at the moment in the node.js code.

sammachin:
@davehorton I've updated this PR to disable the node based cleanup entirely (commented out rather than removed for now)
We can then rely soley on the new cron script in deployment tools

In addition in #1301 I've fixed the issue that was creating a lot of the leftover temp files so the only reason this cleaup would should now be needed is in the event of a crash.

davehorton:
Only concern is that we have K8S clusters out there that may actually rely on this functionality

sammachin:
@davehorton do you think we should remove the OS cronjob then and put the fixes into this one to find all the relevent files?
OR do we want to have the same functionality in 2 places with the os cron and this function?

PR #1301, opened 2025-07-25
use tmpFiles list of parent call-session

### LINKED ISSUE #1299: temp audio files in a ConfirmCallSession are not cleaned up
When a verb like enqueue or conference creates a ConfirmCallSession in order to execute say verbs within something like a waitHook that is a new callsession and as such has its own tmpfiles set for the temporary audio files that are played.

However ConfirmCallSession has overidden the _clearResources function of CallSession and this contains the code to cleanup those audio files.

Have tried adding the code to the overidden function but it doesn't execute, need to investifate further.

sammachin:
The ConfirmCallSession will now use the tmpFiles set of the parent CallSession to store the references to the files created, this means that when the call ends these files are cleaned up by the CallSession _clearRescources funciton

fixes #1299

### DIFF
--- a/lib/session/confirm-call-session.js
+++ b/lib/session/confirm-call-session.js
@@ -8,7 +8,8 @@ const CallSession = require('./call-session');
 
  */
 class ConfirmCallSession extends CallSession {
-  constructor({logger, application, dlg, ep, tasks, callInfo, accountInfo, memberId, confName, rootSpan, req}) {
+  // eslint-disable-next-line max-len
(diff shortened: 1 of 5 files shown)


davehorton:
It seems to me that this works slightly differently than described above, from the code I would say "ConfirmCallSession will now use the tmpFiles set of the _superclass_ CallSession", rather than the "parent" CallSession.  If possible, maintaining the files until the parent call session ended would indeed be preferable I think

PR #1304, opened 2025-07-30
fix assert, and force methods to upper case

sammachin:
Fixes a bug where the assertat L122 was not having any effect therefore invalid methods were being actioned to the callHook url resulting in 400 errors being returned.

Also added function to set the hook.method to uppercase so if the user enters `post` it will still work and not require them to enter `POST`


sammachin:
added an alert on updateCall errors as the API response is just a 202 accepted before command is validated 
image

PR #1325, opened 2025-08-19
fixed say verb store audio to cache with wrong hash

xquanluu:
https://jambonz.freshdesk.com/a/tickets/1198


lib/tasks/say.js
```diff
@@ -326,6 +329,8 @@ class TaskSay extends TtsTask {
                 //2xx response codes indicate success
                 199 < response_code && response_code < 300 &&
                 evt.variable_tts_cache_filename &&
+                // only store if playback IDs match
```
rammohan-y, line 332: dont we need to compare playbackid and evt.variable_tts_playback_id?

PR #1327, opened 2025-08-20
Fix for say verb caching race condition

rammohan-y:
Second solution to fix say verb race condition issue


davehorton (approved):

davehorton:
hmm please resolve conflicts

Issue #1346, opened 2025-09-04
uncaught exception on jambonz.cloud


davehorton:
```
{"level":20,"time":1756996723040,"pid":1261,"hostname":"ip-172-20-10-193","msg":"rest:dial is being killed"}
AssertionError [ERR_ASSERTION]: provided socket has been destroyed
    at DrachtioAgent._makeRequest (/home/admin/apps/jambonz-feature-server/node_modules/drachtio-srf/lib/drachtio-agent.js:278:12)
    at DrachtioAgent.request (/home/admin/apps/jambonz-feature-server/node_modules/drachtio-srf/lib/drachtio-agent.js:333:17)
    at Request.cancel (/home/admin/apps/jambonz-feature-server/node_modules/drachtio-srf/lib/request.js:83:17)
    at TaskRestDial.kill (/home/admin/apps/jambonz-feature-server/lib/tasks/rest_dial.js:60:16)
    at listOnTimeout (node:internal/timers:581:17)
    at process.processTimers (node:internal/timers:519:7) {
  generatedMessage: false,
  code: 'ERR_ASSERTION',
  actual: false,
  expected: true,
  operator: '=='
}
{"level":50,"time":1756996723127,"pid":1261,"hostname":"ip-172-20-10-193","msg":"Uncaught exception - process will restart","app_name":"jambonz-feature-server","error":"provided socket has been destroyed","stack":"AssertionError [ERR_ASSERTION]: provided socket has been destroyed\n    at DrachtioAgent._makeRequest (/home/admin/apps/jambonz-feature-server/node_modules/drachtio-srf/lib/drachtio-agent.js:278:12)\n    at DrachtioAgent.request (/home/admin/apps/jambonz-feature-server/node_modules/drachtio-srf/lib/drachtio-agent.js:333:17)\n    at Request.cancel (/home/admin/apps/jambonz-feature-server/node_modules/drachtio-srf/lib/request.js:83:17)\n    at TaskRestDial.kill (/home/admin/apps/jambonz-feature-server/lib/tasks/rest_dial.js:60:16)\n    at listOnTimeout (node:internal/timers:581:17)\n    at process.processTimers (node:internal/timers:519:7)","pid":1261,"timestamp":"2025-09-04T14:38:43.127Z"}
```

xquanluu:
closing this ticket as PR merged

PR #1347, opened 2025-09-05
update drachtio-srf version

xquanluu:
#1346


davehorton:
@xquanluu please see test failure

PR #1349, opened 2025-09-05
Fix/1345

sammachin:
stops dial and createCall from trying to figure out a best fit trunk if its not specified where the account has LCR setup


sammachin:
waiting on https://github.com/jambonz/db-helpers/pull/46 to be merged and published so I can update dependency

PR #1383, opened 2025-10-07
if call is transferred from another FS then always answer

### LINKED ISSUE #1380: transfer outbound conference participant between feature servers fails
When running with multiple FS and an  OUTBOUND call participant is initially created on a different FS to where the conference is hosted the call fails when the participant is transferred to the new FS, the call then hangs.
this is the error on the recieving FS
```
[11:47:54.287] ERROR (3392625): Error executing task  #0:1: conference
    callId: "c110e8a2-1d44-123f-61a3-0220133cdfa7"
    callSid: "d5bdef2e-2b1b-4d99-ba5c-39248dc176c6"
    accountSid: "c322820d-6250-41b0-9129-7929405e52ed"
    callingNumber: ""
    calledNumber: "context-28467205-32b2-4024-9e9a-9c01888a18a7"
    traceId: "ba7846d44b34ce4bbc82cc9b05df47e4"
    err: {
      "type": "TypeError",
      "message": "Cannot set properties of undefined (setting 'connectTime')",
      "stack":
          TypeError: Cannot set properties of undefined (setting 'connectTime')
              at Conference.exec (/home/admin/apps/jambonz-feature-server/lib/tasks/conference.js:84:43)
              at async InboundCallSession.exec (/home/admin/apps/jambonz-feature-server/lib/session/call-session.js:1287:11)
    }
```

sammachin:
Currently if the call being transferred was originally an outbound call then the direction thats retrieved from redis is outbound and the invite of the refer from the other FS is never answered, However a transferredCall will always need to be answered regardless of CallDirection.

Fixes #1380

### DIFF
--- a/lib/session/call-session.js
+++ b/lib/session/call-session.js
@@ -2399,7 +2399,7 @@ Duration=${duration} `
         this.logger.debug(`endpoint was destroyed!! ${this.ep.uuid}`);
       });
 
-      if (this.direction === CallDirection.Inbound) {
+      if (this.direction === CallDirection.Inbound || this.application?.transferredCall) {
         if (task.earlyMedia && !this.req.finalResponseSent) {
           this.res.send(183, {body: ep.local.sdp});
           return {ep};


sammachin:
Tested inbound and outbound legs being added to the conference both from the same FS and from different FS.

davehorton (approved):

Issue #1200, opened 2025-05-16
Multiple gather with nested play streams not playing


rammohan-y:
In a multiple gather nested play scenarios using streaming, sometimes the stream sent from second gather->play onwards is not playing

Sample code below:

```
const http = require('http')
const express = require('express')
const port = process.env.WS_PORT || 3011;
const port2 = process.env.WS_PORT2 || 3021;
const bodyParser = require('body-parser')
const OpenAI = require("openai");
const OPENAI_API_KEY = '<OPEN API KEY>'; // replace with your API key
const openai = new OpenAI({ apiKey: OPENAI_API_KEY });
const expressApp = express();
const server = http.createServer(expressApp)
expressApp.use(bodyParser.json())


const app = express();
app.use(bodyParser.json())

app.listen(port2,
  function (err) {
    if (err) console.log(err);
    console.log("Server listening on PORT", port2);
  });
expressApp.post('/register',
  (req, res) => {
    console.log("inside register")
    let msg = {
      'status': 'ok',
    };
    res.end(JSON.stringify(msg));
  });
expressApp.get('/register',
  (req, res) => {
    let msg = {
      'status': 'ok',
    };
    res.end(JSON.stringify(msg));
  });
server.listen(port, function () {
  console.log(`**** SAVG Listener Server running on port ${port} **** `)
})


const WebSocket = require('ws')
const wss = new WebSocket.Server({ server, path: "/socket" });
const { WebhookResponse } = require('@jambonz/node-client');

function ack(ws, msgid, res) {
  if (res) {
    res = res instanceof WebhookResponse ? res.toJSON() : res;
  }
  let msg = {
    type: 'ack',
    msgid
  };
  if (res) msg = { ...msg, data: res };
  ws.send(JSON.stringify(msg));
}
function sendCommand(ws, payload, queueCommand = false) {
  payload = payload instanceof WebhookResponse ? payload.toJSON() : payload;

  const msg = {
    type: 'command',
    command: 'redirect',
    queueCommand,
    data: payload
  };
  ws.send(JSON.stringify(msg));
}
function sendTtsTokens(ws, text, queueCommand = false) {
    let msg = {
      type: 'command',
      command: 'tts:tokens',
      queueCommand,
      data: {
        id: new Date().getTime(),
        tokens: text
      }
    }
    ws.send(JSON.stringify(msg));
}
function flushTtsTokens(ws, queueCommand = false) {
  let msg = {
    type: 'command',
[...]

PR #1412, opened 2025-10-28
fix say verb does not close streaming when finish say

xquanluu:
#1200

### DIFF
--- a/lib/tasks/tts-task.js
+++ b/lib/tasks/tts-task.js
@@ -85,55 +85,66 @@ class TtsTask extends Task {
   }
 
   async setTtsStreamingChannelVars(vendor, language, voice, credentials, ep) {
-    const {api_key, model_id, api_uri, custom_tts_streaming_url, auth_token} = credentials;
-    let obj;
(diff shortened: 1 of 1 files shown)


xquanluu:
Hi @davehorton @vdharashive 

I have fixed this issue and it's working with Ram js application.

the issue is that:
1/ when gather finished, it does not closed nested streaming say verb.
2/ when nested stream say verb is finished, ttsStreamingBuffer class remove all streaming listeners but have not reset his own eventHandlers list properties, causing next nested say stream can open new streaming but have not been able to detect connection success event from freeswitch module.

I have tested this on jambonz.me and it's working fine.

vdharashive:
@xquanluu when are we making connection to tts engine ? are we making at the time of doflush or when say command start execution , what i noticed in multiple gather with stream : true doflush alway goes in no connection open and then timer set at 1 sec later will flush it

xquanluu:
it's done @davehorton

davehorton (approved):

Issue #1248, opened 2025-06-20
Fallback ASR not working for second endpoint


rammohan-y:
In a agent transfer scenario (dial verb), when the primary ASR fails, the fallback ASR is not working on any one of the endpoint, 

- Most of the times the agent leg doesn't fallback but user leg fallsback
- Some times occassionaly, user leg doesn't fallback but agent leg fallsback
- Very rarely both the legs doesn't fallback

vdharashive:
@rammohan-y can we have PR for this

vdharashive:
@davehorton see if @xquanluu can look into this

PR #1427, opened 2025-11-10
Fix 1248 fallback issue

rammohan-y:
Fix for issue https://github.com/jambonz/jambonz-feature-server/issues/1248


davehorton:
There are a couple of things I don't love here.
1. You've changed the signature of `_initFallback` to take an endpoint, but you dont pass an endpoint when you call it from gather or transcribe.  I see that it works but I think it increases the complexity of understanding the code a bit
2. You've added a new property to endpoint `fallbackInitiated` and I am not yet sure whether that is needed; again it increases complexity.

I am going to continue to review this PR in more detail and work to recreate the test cases to make sure we have a solution that is simple and easy to understand

rammohan-y:
@davehorton can you help with this issue to be fixed

@vdharashive @oddsix @xquanluu

PR #1429, opened 2025-11-11
Add ability to override certain tts streaming options via the config …

oddsix:
Allow streaming tts settings to be overridden via the options sub-element of the config verb.

### DIFF
--- a/lib/tasks/tts-task.js
+++ b/lib/tasks/tts-task.js
@@ -85,25 +85,20 @@ class TtsTask extends Task {
   }
 
   async setTtsStreamingChannelVars(vendor, language, voice, credentials, ep) {
-    const {api_key, model_id, api_uri, custom_tts_streaming_url, auth_token} = credentials;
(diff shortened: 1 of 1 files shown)


lib/tasks/tts-task.js
```diff
@@ -86,34 +86,50 @@ class TtsTask extends Task {
 
   async setTtsStreamingChannelVars(vendor, language, voice, credentials, ep) {
     const {api_key, model_id, api_uri, custom_tts_streaming_url, auth_token} = credentials;
+    // api_key, model_id, api_uri, custom_tts_streaming_url, and auth_token are encoded in the credentials
+    // allow them to be overriden via config, using options
+    const local_api_key = (this.options.api_key !== null &&
```
davehorton: can we change to this format
```
const local_api_key = this.options.api_key ?? api_key;
```

davehorton (approved):

PR #1440, opened 2025-11-18
fix error when tts stream invoked without config

oddsix:


oddsix:
@davehorton comments have been incoporated into PR.

vdharashive:
@oddsix the main issue of having config verb to use  option properties still not working

PR #1444, opened 2025-11-21
Compare sdp to determine if transcoding is being used.

oddsix:


lib/tasks/dial.js
```diff
@@ -468,6 +471,52 @@ class TaskDial extends Task {
     this.dials.clear();
   }
 
+  /**
+   * Check if two SDP endpoints are using the same primary audio codec
+   * @param {object} ep1 - First endpoint object
+   * @param {object} ep2 - Second endpoint object
+   * Sets this.mediaNotSame to false if same codec (no transcoding needed), true otherwise
+   */
+  _checkTranscoding(ep1, ep2) {
```
davehorton: lets create a new file utils/sdp-utils.js and place a function there `getLeadingCodec(sdp)` which parses an SDP and provides the first audio codec in the list

lib/tasks/dial.js
```diff
@@ -979,7 +932,11 @@ class TaskDial extends Task {
       }
     }
     /* basic determination to see if call is being transcoded */
-    this._checkTranscoding(this.epOther, sd.ep);
+    const codecA = getLeadingCodec(this.ep.local.sdp);
```
davehorton: this should be
```
const codecA = getLeadingCodec(this.epOther.local.sdp);
```

PR #1446, opened 2025-11-23
SoundHound support audio endpoint from speech credential

xquanluu:
https://jambonz.freshdesk.com/a/tickets/1655


davehorton:
can you rebase

PR #1468, opened 2025-12-10
add label to STT/TTS alerts

sammachin:
Add the label to the alert message for TTS/STT not configured, 
Setting the label on a speech provider Then not using that lable in the verb is a common user error, by adding the label (or None) to the alert message this should prompt users to check the label config.
Example alert messages for both the label set in the verb and not configured (TestLabel) and where the provider has a label but not set in the verb (none)
Screen Shot 2025-12-10 at 5 35 00 PM
Created as draft as this also requires an update to @jambonz/time-series first


sammachin:
waiting on https://github.com/jambonz/time-series/pull/18

PR #1473, opened 2025-12-16
listenDuringPrompt to true gather timer not set and empty transcript not to extend timer

vdharashive:
- When listenDuringPrompt is true, transcription is active during the prompt and start the timer
- Timer now starts once at gather begin and continues until timeout or input
- Prevents multiple timer starts for the same gather operation for listenDuringPrompt is true and timer already started
- Update emptyTranscript detection to use .trim() instead of exact empty string check
- This ensures transcripts containing only whitespace are treated as empty
- Prevents whitespace-only transcripts from incorrectly resetting gather timers
- Applies to both final and non-final transcripts


vdharashive:
@xquanluu can you review this if you need any help please let me know

lib/tasks/gather.js
```diff
@@ -1090,7 +1094,7 @@ class TaskGather extends SttTask {
         }
       }
 
-      if (!emptyTranscript) {
+      if (!emptyTranscript && evt.alternatives[0].transcript.trim() !== '') {
```
xquanluu: if we ignore like this.
Can you please set emptyTranscript = evt.alternatives[0].transcript.trim() === '' as well.
we cannot have emptyTranscript = true, but evt.alternatives[0].transcript.trim() === '', this will break system.
let make consisten here.

PR #1477, opened 2025-12-17
fixed: should start transcription and timeout timer when listenDuringPrompt=true

xquanluu:
…


vdharashive:
@xquanluu and @davehorton please do not merge this we are doing some more testing with all different cases

vdharashive:
Hi @xquanluu  i have update the ticket https://jambonz.freshdesk.com/support/tickets/1783 with another bug for this fix

PR #1489, opened 2026-01-09
freeswitch capture sip error and propagate  the same error

vdharashive:
fix: propagate SIP  error to SBC on endpoint allocation failure
    
    When FreeSWITCH returns a SIP 488 'Not Acceptable Here' error during
    endpoint allocation (e.g., codec incompatibility), this error was not
    being propagated back to the SBC/client. Instead, the call would wait
    indefinitely for websocket commands or return a generic 603 response.
    
    Implementation:
    - In _evalEndpointPrecondition(), detect SipError by checking
      err.type === 'SipError' or err.name === 'SipError'
    - Extract the SIP status code (e.g., 488), reason, and the Reason
      header from the error response (e.g., Q.850;cause=88;text=INCOMPATIBLE_DESTINATION)
    - Send the SIP error response immediately to the SBC with:
      - X-Reason header: endpoint allocation failure details
      - Reason header: original Q.850 cause from FreeSWITCH
    - Notify call status change as Failed with proper SIP status
    - Release the call immediately instead of waiting for commands
    
    Also added fallback handling in InboundCallSession._onTasksDone() to
    propagate the stored error if immediate send was not possible.

### DIFF
--- a/lib/session/call-session.js
+++ b/lib/session/call-session.js
@@ -2501,7 +2501,7 @@ Duration=${duration} `
       else {
         this.logger.error(err, `Error attempting to allocate endpoint for for task ${task.name}`);
         // Check for SipError type (e.g., 488 codec incompatibility)
-        const isSipError = err.type === 'SipError' || err.name === 'SipError';
+        const isSipError = err.name === 'SipError';
         if (isSipError && err.status) {
           // Extract Reason header from SIP response if available (e.g., Q.850;cause=88;text="INCOMPATIBLE_DESTINATION")
           const sipReasonHeader = err.res?.msg?.headers?.reason;


lib/session/call-session.js
```diff
@@ -2500,6 +2500,36 @@ Duration=${duration} `
       }
       else {
         this.logger.error(err, `Error attempting to allocate endpoint for for task ${task.name}`);
+        // Check for SipError type (e.g., 488 codec incompatibility)
+        const isSipError = err.type === 'SipError' || err.name === 'SipError';
```
davehorton: please make this check just against err.name.  There is no need to check both

vdharashive: @davehorton changes done

PR #1500, opened 2026-01-21
add statusHook to redirect verb

### LINKED ISSUE #1492: Feat: Redirect verb should allow updating statusHook
Currently the redirect verb only updates the actionHook so if transferring the call to a new applicaiton the statusHook messages are still sent to the original server configured in the app.

We need to allow a new optional field in the redirect verb to set the statusHook.

actionHook should be required and always set
statusHook is optional, if not set the original url is maintained

sammachin:
adds support for statusHook as an optional param on a redirect verb,
tested with, absolute and relative urls for both actionHook and actionHook & statusHook
tested with websocket redirects

fixes #1492

I also removed the `const URL = require('url');` line as since #1488 this was failing with `TypeError: URL is not a constructor`
Could have also wrapped URL in {} but consulting with claude just removing the import is cleaner as URL has been global since Node10. 

Left as draft awaiting https://github.com/jambonz/verb-specifications/pull/120 tp be merged and published to update verb specs dependency

### DIFF
--- a/lib/utils/action-hook-delay.js
+++ b/lib/utils/action-hook-delay.js
@@ -118,6 +118,13 @@ class ActionHookDelayProcessor extends Emitter {
     this.logger.debug('ActionHookDelayProcessor#_onNoResponseTimer');
     this._noResponseTimer = null;
 
+    /* check if endpoint is still available (call may have ended) */
+    if (!this.ep) {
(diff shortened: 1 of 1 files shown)


davehorton:
can you rebase?

PR #1524, opened 2026-03-11
add deleteCall function to properly cleanup Ultravox session

oddsix:
Call delete REST API to properly clean up session.

### DIFF
--- a/lib/tasks/llm/llms/ultravox_s2s.js
+++ b/lib/tasks/llm/llms/ultravox_s2s.js
@@ -146,6 +146,23 @@ class TaskLlmUltravox_S2S extends Task {
     return data;
   }
 
+  async deleteCall() {
+    if (!this.callId) return;
+    const baseUrl = 'https://api.ultravox.ai';
+    const url = `${baseUrl}/api/calls/${this.callId}`;
+    try {
+      const {statusCode} = await request(url, {
+        method: 'DELETE',
+        headers: {
+          'X-API-Key': this.apiKey
+        }
+      });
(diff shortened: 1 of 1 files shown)


lib/tasks/llm/llms/ultravox_s2s.js
```diff
@@ -200,6 +220,8 @@ class TaskLlmUltravox_S2S extends Task {
     this._api(cs.ep, [cs.ep.uuid, SessionDelete])
       .catch((err) => this.logger.info({err}, 'TaskLlmUltravox_S2S:kill - error deleting session'));
 
+    this.deleteCall();
```
davehorton, line 223: hmm, can you tell me what the SessionDelete message above this does?  Are we sure this is not redundant?

oddsix: That tells the freeeswitch module to tear down the WS connection to Ultravox, which doesn't call the delete api endpoint.

davehorton: ok what specifically does the SessionDelete client message do?  Can you point me to the ultravox docs for that as well as the new api you added

oddsix: The delete API call can be found here: https://docs.ultravox.ai/api-reference/calls/overview#deleting-calls  The problem being reported is during abrupt call termination. The next call using the agent id picks up the memory of the previous call, so the context is off and potentially sensitive data is shared.

PR #1525, opened 2026-03-15
Append sip realm for rest dial

avoylenko:


rhondahollis:
@avoylenko The logic all looks good. I would just add a safety check for an undefined sip_realm:

```
if (!targetName.includes('@')) {
            if (!account.sip_realm) throw new Error(`cannot dial user ${targetName}: 
              account has no sip_realm configured`);
            logger.debug(`appending sip realm ${account.sip_realm} to target name ${targetName}`);
            targetName = `${targetName}@${account.sip_realm}`;
          }
          uri = `sip:${targetName}`;
          to = targetName;
```

avoylenko:
@rhondahollis agreed, but in that case we are throwing `500` status code to the client, what about throwing `DbErrorBadRequest() `error?

rhondahollis:
> DbErrorBadRequest

@avoylenko Yes, you can use DbErrorBadRequest here so that a 400 is returned instead.

avoylenko:
@rhondahollis pushed

rhondahollis (approved):

davehorton (approved):

PR #1529, opened 2026-03-24
ensure sbcCallId is added to callInfo

rhondahollis:
If no provisional response is received, the sbc_callid was not being added to the callInfo.

https://jambonz.freshdesk.com/a/tickets/2133


rhondahollis:
@davehorton I was able to confirm the fix with a call that only receives a 200 OK.

Screenshot 2026-03-25 at 1 41 23 PM

davehorton (approved):

Issue #1505, opened 2026-01-29
no application_sid in CDR when call isn't answered


sammachin:
If the application is something simple like say a Dial verb that tries to connect the call onto another endpoint, and that dial verb fails because the far end rejects the call then in the call details on recent calls no application_sid is shown.
It looks as if this value is only set when the call is answered. (Putting an answer before the dial sets the sid)

In addition the jaeger tracing is also blank.

<img width="1245" height="1033" alt="Image" src="https://github.com/user-attachments/assets/6c2966ae-6ae4-4e5a-8473-e004d5660b18" />

This is confusing for users as they think that there is an issue routing to the application rather than looking at the issues with the dial.

rhondahollis:
@sammachin Please see PRs:
https://github.com/jambonz/jambonz-feature-server/pull/1531
https://github.com/jambonz/sbc-inbound/pull/240

davehorton:
I don't love the use of redis to pass information between sbc-inbound and jambonz-feature-server.  That is not a pattern we have used before and introduces additional coupling and potential race conditions.  Not sure if this issue is even worth fixing if the cost is the additional complexity.  Lets discuss offline the impact of this bug, and which customers it is impacting

rhondahollis:
@sammachin @davehorton Here is a summary of root cause for the issue and the potential fixes that I've been looking at. In my testing of this, I've been calling into a test application that has a simple dial verb that routes to a SIPP app which then responds with a 603.

### Root cause: timing race in the failure-CDR write path
sbc-inbound writes failure CDRs from a res.once('end') listener in [middleware.js](https://github.com/jambonz/sbc-inbound/blob/main/lib/middleware.js#L61). That listener serializes req.locals.cdr to the writeCdrs payload at the moment 'end' fires — and crucially, the {...req.locals.cdr, ...} object spread is evaluated synchronously at call time, not when the writeCdrs promise resolves. Whatever isn't on req.locals.cdr by then never reaches InfluxDB.

[...]

PR #1531, opened 2026-03-26
ensure traceId is set on CallRecord

rhondahollis:
Fix for https://github.com/jambonz/jambonz-feature-server/issues/1505
Corresponding change in sbc-inbound: https://github.com/jambonz/sbc-inbound/pull/240

These changes ensure an X-Trace-Id is added to the error responses sent to sbc-inbound. 
I also added a Redis entry with the traceId keyed to the callId, so that there is still a traceId on Call Records when either the user hangs up before the call is answered or the application doesn't provide an actionHook to handle a failed dial, thereby resulting in the caller hanging up due to lack of a final response making it back to sbc-inbound.

### DIFF
--- a/lib/middleware.js
+++ b/lib/middleware.js
@@ -28,6 +28,7 @@ module.exports = function(srf, logger) {
     registrar,
     lookupClientByAccountAndUsername
   }  = srf.locals.dbHelpers;
+  const {addKey} = srf.locals.dbHelpers;
   const {
     writeAlerts,
     AlertType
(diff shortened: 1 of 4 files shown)


lib/middleware.js
```diff
@@ -142,6 +143,14 @@ module.exports = function(srf, logger) {
...
     };
 
+    /* write a short-lived Redis key mapping the original Call-ID (X-CID) to the traceId,
+       so sbc-inbound can look it up for CDRs even when a CANCEL is sent to sbc-inbound */
+    const externalCallId = req.get('X-CID');
+    if (externalCallId) {
+      addKey(`callid:${externalCallId}`, traceId, 300)
+        .catch((err) => logger.error(err, 'createRootSpan: error writing traceId to Redis'));
+    }
+
```
rhondahollis: @davehorton @sammachin Theses changes in middleware.js to add the Redis call are particular to handling scenarios where the final response is not propagated back to sbc_inbound: 
- caller hangs up before receiving a final response
- no actionHook on a dial verb, which results in a timeout and feature-server generating a CANCEL. This is what I kept hitting in my own test scenario, but I don't know frequent this would be for customers. 

If we decide to remove just these changes (and the corresponding one in sbc-inbound), the changes in the other files on this PR would still ensure application and trace ID are added to the Call Record in the scenarios where a final response is returned to sbc-inbound.

PR #1534, opened 2026-03-27
add null check for eventHook

rhondahollis:
Fixes TypeError when no eventHook was defined on the Llm verb.

### DIFF
--- a/lib/tasks/llm/llms/ultravox_s2s.js
+++ b/lib/tasks/llm/llms/ultravox_s2s.js
@@ -286,6 +286,10 @@ class TaskLlmUltravox_S2S extends Task {
   }
 
   _sendLlmEvent(type, evt) {
+    if (!this.eventHook) {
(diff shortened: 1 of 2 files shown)


lib/tasks/llm/llms/ultravox_s2s.js
```diff
@@ -286,6 +286,10 @@ class TaskLlmUltravox_S2S extends Task {
   }
 
   _sendLlmEvent(type, evt) {
+    if (!this.eventHook) {
+      this.logger.info({evt}, 'TaskLlmUltravox_S2S:_sendLlmEvent - no eventHook defined!');
```
xquanluu: In my humble opinion,

Please help to use debug level, info will make a lot of logs printed.
this is not important place to log as well, might be return is good enough.

davehorton: agree

PR #1536, opened 2026-04-06
add evt check in mediaTimeoutHandler

rhondahollis:
Added a check for the event reason to make sure it actually is a media timeout, and added null check for endpoint. Support ticket #2170 identified a scenario where a broken pipe on the FreeSWITCH connection was ending up in the media timeout handler and tearing down the call.


xquanluu:
look good to me.

Issue #1538, opened 2026-04-17
allow create outbound call rest api on tls/srtp


xquanluu:
Allow rest call API to create outbound call with TLS/SRTP without a need of carrier sip gateway configuration

rhondahollis:
@xquanluu I've added PRs for this change:
https://github.com/jambonz/sbc-outbound/pull/222
https://github.com/jambonz/jambonz-api-server/pull/550
https://github.com/jambonz/jambonz-feature-server/pull/1539

PR #1539, opened 2026-04-17
Support srtp param on CreateCall API

rhondahollis:
https://github.com/jambonz/jambonz-feature-server/issues/1538

Provides support for an srtp param on the Create Call API.

Goes with updates in sbc-outbound and jambonz-api-server:
https://github.com/jambonz/sbc-outbound/pull/222
https://github.com/jambonz/jambonz-api-server/pull/550

### DIFF
--- a/lib/http-routes/api/create-call.js
+++ b/lib/http-routes/api/create-call.js
@@ -58,6 +58,10 @@ router.post('/',
       logger.debug({ err }, `invalid app_json: ${err.message}`);
     }
 
+    // capture and remove srtp from body before verb validation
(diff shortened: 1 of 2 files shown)


lib/http-routes/api/create-call.js
```diff
@@ -58,6 +58,10 @@ router.post('/',
       logger.debug({ err }, `invalid app_json: ${err.message}`);
     }
 
+    // capture and remove srtp from body before verb validation
```
rhondahollis, line 61: Should we add the srtp parameter to the dial verb as well?

PR #1560, opened 2026-07-09
feat: surface in-dialog hold/un-hold to the application

joeheung:
## What & why

Applications currently have no way to observe a mid-call hold/un-hold. The feature-server handles the in-dialog re-INVITE internally (re-modifies the endpoint, returns 200) but never notifies the app, so a transcription or agent-assist app can't pause/resume when a caller is held or retrieved.

This detects the hold state transition on a re-INVITE and delivers it over the existing `sipRequestWithinDialogHook` channel as a `verb:hook`, plus emits a local `hold`/`unhold` event so in-process tasks (e.g. transcribe) can react.

## How

- **`call-session.js`** — ordinary calls: reuse `isOnhold()` (`a=sendonly`/`a=inactive`).
- **`siprec-call-session.js`** — SIPREC overrides `_onReinvite`, so a separate handler is required. SIPREC recording streams are sendonly by nature, so a stream flipping to `a=inactive` is treated as the hold signal; the full multipart body (SDP + rs-metadata) is always forwarded. Fires the local event on transition only; other re-INVITEs are delivered as `event:"reinvite"`.

Delivery uses `requestor.request('verb:hook', ...)` — the same channel `_handleRefer` uses — rather than a task-bound `performHook`, so events are not dropped if the current task has ended.

Enable per call with:
```json
{"verb":"config","sipRequestWithinDialogHook":"/hold-events"}
```

## Testing

- Added `test/hold-unhold-test.js` (35 assertions) covering hold/unhold transition detection, transition-only firing, local event emission, and hook payload shape for both ordinary calls and SIPREC. Wired into `test/index.js`.
- `npm run jslint` passes.
- Confirmed live against Cisco CUBE (Cisco-SIPGateway/IOS-17.12.7b) over SIPREC: both recording streams flip sendonly<->inactive on hold/resume, each transition emitted correctly.


davehorton:
Can you rebase on latest main?
Can you guard this new feature with a new env var that must be set in order to see these events, call it `JAMBONES_HOLD_UNHOLD_EVENTS`

lib/session/call-session.js
```diff
@@ -2810,6 +2812,43 @@ Duration=${duration} `
...
   }
 
+  /**
+   * Detect a hold/un-hold transition on an in-dialog re-INVITE and notify the
+   * application (reusing the sipRequestWithinDialogHook channel). Also emits a
+   * local 'hold'/'unhold' event so in-process tasks (e.g. transcribe) can react.
+   * Opt-in: only runs when JAMBONES_HOLD_UNHOLD_EVENTS is set.
+   */
+  _notifyHoldState(req) {
+    if (!process.env.JAMBONES_HOLD_UNHOLD_EVENTS) return;
```
davehorton: actually please look at config.js where we centralize env vars, add it there using existing pattern

davehorton (changes requested):

davehorton (approved):

davehorton:
oh darn.  @joeheung see this failure
https://github.com/jambonz/jambonz-feature-server/actions/runs/29021522305/job/86129615160

please give another PR that fixes it.  Did you add test cases but not run them?

Issue #178, opened 2022-10-21
dial: ability to set From host to specific value


davehorton:
Some carriers or PBX's require that when sending them an INVITE the From header should not have the IP of the sending server but rather a specific domain (e.g. blf.finotel.com).

It would be nice to have an option added to the 'target' property of the [dial](https://www.jambonz.org/docs/webhooks/dial/) verb called 'fromHost' or something that allowed the user to optionally specify the host value of the From header in the outgoing INVITE

davehorton:
fixed in main branch

PR #1579, opened 2026-08-20
add sip_reason_header to status Callback- #178

xquanluu:
https://jambonz.freshdesk.com/a/tickets/2811

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,6 +12,7 @@ jobs:
           node-version: 20
       - run: npm ci
       - run: npm run jslint
+      - run: npm run test:unit
       - name: Install Docker Compose
         run: |
           sudo curl -L "https://github.com/docker/compose/releases/download/1.29.2/docker-compose-$(uname -s)-$(uname -m)" -o /usr/local/bin/docker-compose
--- a/lib/http-routes/api/create-call.js
+++ b/lib/http-routes/api/create-call.js
@@ -339,7 +339,7 @@ router.post('/',
             cs.callInfo.sbcCallid = prov.get('X-CID');
             if ([180, 183].includes(prov.status) && prov.body) connectStream(prov.body);
             restDial.emit('callStatus', prov.status, !!prov.body);
-            cs.emit('callStatusChange', {callStatus, sipStatus: prov.status});
+            cs.emit('callStatusChange', {callStatus, sipStatus: prov.status, msg: prov});
           }
         });
         connectStream(dlg.remote.sdp);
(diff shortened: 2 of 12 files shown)


davehorton:
Approach looks right — `msg.get('Reason')` works through `Response`'s delegate, `err.res` is genuinely set by srf on both createUAC failure paths, and clearing unconditionally (rather than the `if (x)` idiom used for `sipStatus`/`sipReason`) is correct so a cause from a 183 doesn't linger onto the 200. Two things:

### 1. The redis stale-value fix doesn't work for `CallSession`

The `|| ''` in `updateCallStatus` and its comment exist so `hmset` overwrites a stale header. That holds for `SingleDialer`, which sends the raw `callInfo` to redis. `CallSession` sends `toJSON()` (`call-session.js:3142`), and `toJSON()` filters `sipReasonHeader` on truthiness — so `''` never reaches `hmset`:

```
after 183, toJSON has: Q.850;cause=31
raw value: ""
toJSON key present (what CallSession sends to redis): false
```

The split is by session class, not by call direction. `place-outdial` is required only by `dial.js`, so `SingleDialer` covers `dial`-verb child calls; inbound, REST-created (own `createUAC` in `create-call.js:273` + `RestCallSession`) and adulting all inherit `CallSession._notifyCallStatusChange` and keep the stale header on `GET /Calls/:sid`.

Fix at the boundary, leaving the webhook payload alone:

```js
this.updateCallStatus({...this.callInfo.toJSON(), sipReasonHeader: this.callInfo.sipReasonHeader || ''},
  this.serviceUrl)
```

The unit test asserts against the raw instance, so it passes while the real path is broken — it should go through `toJSON()` for the CallSession case.

### 2. The caller-abandoned race path misses the CANCEL

`middleware.js:120` has the CANCEL in hand and discards it (`req.once('cancel', (sipMsg) => { req.locals.canceled = true; })`), so the constructor's `this._onCancel()` at `inbound-call-session.js:28` passes nothing. The header lands or doesn't depending on whether the CANCEL beat the application fetch.

This is the weaker of the two — a consistency hole rather than a bug, since wiring `_onCancel(cancelReq)` already commits to capturing CANCEL. But it's the path where the header is worth the most: the 487 is one we generate ourselves, `'Request Terminated'` is hardcoded, so every abandoned inbound call currently looks identical. A `Reason: SIP;cause=200;text="Call completed elsewhere"` on the CANCEL is the difference between "another branch answered" and "caller gave up", and both are `no-answer` today.

Two lines (`req.locals.cancelReq = sipMsg`, then pass it) — but worth first confirming sbc-inbound relays `Reason` when it proxies the CANCEL, otherwise the fix is a no-op and the change belongs there instead.

davehorton (approved):

PR #1580, opened 2026-08-23
fix: fire sip:refer actionHook when the far end BYEs before any NOTIFY (#1258)

### LINKED ISSUE #1258: sip:refer actionHook not always fired
In the scenario of a sip:refer being sent where the remote end sends back a 202 and then immediatly a Bye the actionHook isn't called, at least in a websocket as the BYE closes the connection before the 15sec timer at https://github.com/jambonz/jambonz-feature-server/blob/f18b62e165bb567cbce3887bc0ea036c6c38e970/lib/tasks/sip_refer.js#L51-L56 completes.

Not sure about webhooks.

Will create an app to re-create

ousamabenyounes:
## Summary

Fix #1258

When a `sip:refer` verb gets a `202` to its REFER it waits for a NOTIFY carrying the final status
of the referred call. The actionHook was only fired from the 15 s timeout callback, so when the far
end sends `BYE` right after the `202` — the exact scenario in the issue — the call session kills the
task, `awaitTaskDone()` resolves, and `exec()` returned **without ever calling `performAction()`**.
The application never sees the actionHook. With a websocket application the connection is then torn
down by `CallSession._clearResources()`, so nothing arrives later either.

- `exec()` now performs the action once after `awaitTaskDone()`, i.e. on every exit path (final
  NOTIFY, NOTIFY timeout, or the task being killed by an early BYE) and while the requestor is
  still up. This is the same shape the `dial` verb already uses (`lib/tasks/dial.js`).
- `_performReferAction()` memoises the action promise, so the hook fires exactly once and `exec()`
  cannot return while a hook started by the NOTIFY path is still in flight.
- `kill()` records its `KillReason`; when the application itself replaced this verb
  (`KillReason.Replaced`), the hook is still sent but its response no longer replaces the
  application a second time — matching `lib/tasks/dial.js:249` and `lib/tasks/enqueue.js:115`.
- The `15000` and `202` literals in this file became named constants while touching those lines.

No change to the payload the application receives on the paths that already worked.

## Test verification (RED → GREEN)

New `test/unit/sip-refer-action-hook.test.js` (7 cases, `node:test`, run by `npm run test:unit`).

**RED — on unmodified `main` with only the new test file added** (`git worktree add --detach <tmp> origin/main`):

```
not ok 1 - sip:refer actionHook fires when the far end sends BYE before any NOTIFY
  ---
  failureType: 'testCodeFailure'
  error: |-
    actionHook should have been called exactly once

    0 !== 1

  code: 'ERR_ASSERTION'
  expected: 1
  actual: 0
  operator: 'strictEqual'
  ...
ok 2 - sip:refer actionHook fires once, with the final status, when a NOTIFY arrives
ok 3 - sip:refer actionHook fires when the far end rejects the REFER
ok 4 - a failing actionHook is logged, not thrown out of the verb
ok 5 - sip:refer actionHook fires when no NOTIFY arrives before the timeout
not ok 6 - exec waits for an in-flight actionHook when a BYE races the final NOTIFY
not ok 7 - an actionHook response replaces the application only when the verb was not itself replaced
# tests 7
# pass 4
# fail 3
```

`not ok 1` is the reported bug: `0 !== 1`, the actionHook was never called. Cases 2–5 pass on `main`
by design — they are regression guards for the paths this diff refactors.

**GREEN — same test file on this branch:**

```
ok 1 - sip:refer actionHook fires when the far end sends BYE before any NOTIFY
ok 2 - sip:refer actionHook fires once, with the final status, when a NOTIFY arrives
ok 3 - sip:refer actionHook fires when the far end rejects the REFER
ok 4 - a failing actionHook is logged, not thrown out of the verb
ok 5 - sip:refer actionHook fires when no NOTIFY arrives before the timeout
ok 6 - exec waits for an in-flight actionHook when a BYE races the final NOTIFY
ok 7 - an actionHook response replaces the application only when the verb was not itself replaced
# tests 7
# pass 7
# fail 0
```

## Full local suite / CI replay

The full local validation suite a fork checkout can actually run — the first three steps of the
`build` job — was replayed in the CI image (`node:20-bookworm-slim`, matching `actions/setup-node` in
`.github/workflows/build.yml`), on `origin/main` first and then on this commit:

| step | `origin/main` baseline | this branch |
|------|------------------------|-------------|
| `npm ci` | ok | ok |
| `npm run jslint` | clean | clean |
| `npm run test:unit` | 8 pass / 0 fail | 15 pass / 0 fail |

Every changed line of `lib/tasks/sip_refer.js` is exercised by the new tests.

The remaining step, the `npm test` regression suite, was **not** replayed locally: per `docs/contributing.md` it needs
your own GCP/AWS speech credentials plus the docker-compose testbed, which a fork checkout does not
have. CI covers it here.

## Follow-up — davehorton review (final NOTIFY → BYE race), commit `d060fb7`

A final NOTIFY (`status >= 200`) whose `eventHook` round trip was still in flight when the far
end BYEd lost `final_referred_call_status` on the actionHook: `kill()` woke `exec()`, which memoised
the action via `_performReferAction({refer_status})` **without** the final status, so the NOTIFY
path's later call carrying it was deduped away. On `main` the app received `final_referred_call_status`
in this case, so this was a regression introduced by the original diff.

Fix: record the status synchronously in `_handleNotify` (`this._finalReferredCallStatus`) **before**
the `eventHook` await, and include it from `exec()` so the memoised action carries it whichever exit
path wins the race. The `200` threshold became `SIP_FINAL_RESPONSE_MIN`.

New 8th case in `test/unit/sip-refer-action-hook.test.js` — *"the final NOTIFY status still reaches the
actionHook when a BYE races the eventHook round trip"*:

- **RED** on `72f9ee0`: `final_referred_call_status must survive a BYE racing the final NOTIFY eventHook — undefined !== 200`
- **GREEN** on `d060fb7`: `# tests 8 # pass 8 # fail 0`; `test:unit` baseline `origin/main` 10 pass → this branch 18 pass, no regressions.

Generated by Ora Studio
Vibe coded by ousamabenyounes

### DIFF
--- a/lib/tasks/sip_refer.js
+++ b/lib/tasks/sip_refer.js
@@ -1,6 +1,12 @@
 const Task = require('./task');
-const {TaskName, TaskPreconditions} = require('../utils/constants');
+const {TaskName, TaskPreconditions, KillReason} = require('../utils/constants');
 const {parseUri} = require('drachtio-srf');
+
+/* how long we wait for a NOTIFY carrying the final status of the referred call */
+const NOTIFY_TIMEOUT_MS = 15000;
+
+/* SIP status returned by a far end that accepted the REFER */
+const REFER_ACCEPTED = 202;
 
 /**
  * sends a sip REFER to transfer the existing call
@@ -15,6 +21,7 @@ class TaskSipRefer extends Task {
     this.referredByDisplayName = this.data.referredByDisplayName;
     this.headers = this.data.headers || {};
     this.eventHook = this.data.eventHook;
+    this._actionPromise = null;
   }
 
   get name() { return TaskName.SipRefer; }
@@ -47,30 +54,30 @@ class TaskSipRefer extends Task {
       this.logger.info(`TaskSipRefer:exec - received ${this.referStatus} to REFER`);
 
       /* if we fail, fall through to next verb.  If success, we should get BYE from far end */
-      if (this.referStatus === 202) {
+      if (this.referStatus === REFER_ACCEPTED) {
         this._notifyTimer = setTimeout(() => {
-          this.logger.info('TaskSipRefer:exec - no NOTIFY received in 15 secs, exiting');
-          this.performAction({refer_status: this.referStatus})
-            .catch((err) => this.logger.error(err, 'TaskSipRefer:exec - error performing action'));
+          this.logger.info(`TaskSipRefer:exec - no NOTIFY received in ${NOTIFY_TIMEOUT_MS} ms, exiting`);
           this.notifyTaskDone();
-        }, 15000);
+        }, NOTIFY_TIMEOUT_MS);
         await this.awaitTaskDone();
         if (this._notifyTimer) {
           clearTimeout(this._notifyTimer);
           this._notifyTimer = null;
         }
       }
-      else {
-        await this.performAction({refer_status: this.referStatus});
-      }
+      /* the far end may send BYE before any NOTIFY arrives, which kills this task while we are
+         awaiting above; performing the action here means the actionHook runs on every exit path -
+         and while the requestor is still up - rather than only when the NOTIFY timer expires */
+      await this._performReferAction({refer_status: this.referStatus});
     } catch (err) {
       this.logger.info({err}, 'TaskSipRefer:exec - error sending REFER');
     }
     this.referSpan?.end();
   }
 
-  async kill(cs) {
+  async kill(cs, reason) {
     super.kill(cs);
+    this.killReason = reason || KillReason.Hangup;
     const {dlg} = cs;
     dlg.off('notify', this.notifyHandler);
     this.notifyTaskDone();
@@ -95,14 +102,27 @@ class TaskSipRefer extends Task {
         }
         if (status >= 200) {
           this.referSpan.setAttributes({'refer.finalNotify': status});
-          await this.performAction({refer_status: 202, final_referred_call_status: status})
-            .catch((err) => {
-              this.logger.error(err, 'TaskSipRefer:exec - error performing action finalNotify');
-            });
+          await this._performReferAction({refer_status: REFER_ACCEPTED, final_referred_call_status: status});
           this.notifyTaskDone();
         }
       }
     }
+  }
+
+  /**
+   * fire the verb's actionHook exactly once, whichever exit path completes the task:
+   * a final NOTIFY, the NOTIFY timeout, or the task being killed by an early BYE
+   */
+  _performReferAction(results) {
+    if (!this._actionPromise) {
+      /* when the app replaced this verb with new commands, still notify it, but do not let the
+         hook response replace the application a second time (same contract as the dial verb) */
+      this._actionPromise = this.performAction(results, this.killReason !== KillReason.Replaced)
+        .catch((err) => this.logger.error({err}, 'TaskSipRefer:_performReferAction - error performing action'));
+    }
+    /* callers await the shared promise, so exec() cannot return - and let the session close the
+       requestor - while an actionHook started by another exit path is still in flight */
+    return this._actionPromise;
   }
 
   _normalizeReferHeaders(cs, dlg) {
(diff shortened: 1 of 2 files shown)


ousamabenyounes:
Local CI replay scope for this PR, so it is clear what was and was not exercised before pushing.

Replayed on `origin/main` (baseline) and again on this commit, inside `node:20-bookworm-slim` to
match `actions/setup-node@v4` in `.github/workflows/build.yml`:

- `npm ci`
- `npm run jslint` — clean on both
- `npm run test:unit` — 8 pass / 0 fail on baseline, 15 pass / 0 fail here (7 new cases, no regressions)

Plus the RED→GREEN pair in the description: the new `test/unit/sip-refer-action-hook.test.js` run
against an unmodified `origin/main` worktree with only that file added (3 of 7 cases fail, including
the reported symptom `0 !== 1` — the actionHook is never called), then against this branch (7/7 pass).

Not replayed locally: the `npm test` regression suite. Per `docs/contributing.md` it needs GCP/AWS
speech credentials and the docker-compose testbed, which a fork checkout does not have. Only files
under `lib/tasks/sip_refer.js` and `test/unit/` are touched, so that suite is the remaining coverage
and CI owns it.

The `CI` workflow run for this PR is currently in `action_required` — first-time-contributor
workflow approval. Happy to rebase or adjust anything if the run turns up something.

davehorton:
One issue I see: a final NOTIFY followed immediately by a BYE loses the `final_referred_call_status` on the eventHook.

In `_handleNotify` the eventHook request is awaited before `_performReferAction` runs:

```js
if (this.eventHook) {
  await cs.requestor.request('verb:hook', this.eventHook, {...});   // round trip
}
if (status >= 200) {
  await this._performReferAction({refer_status: 202, final_referred_call_status: status});
```
NOTIFY(200) → BYE is the normal sequence. But if the BYE arrives during that round trip, `kill()` → `notifyTaskDone()` wakes `exec`, which calls `_performReferAction({refer_status: 202})` first; the NOTIFY handler's call with the final status is then deduped away. On `main` the app received `final_referred_call_status` in this case; with this PR it doesn't.

I'd suggest recording the status synchronously when the final NOTIFY is parsed, and include it from `exec`:


```js
// _handleNotify, before any await
if (status >= 200) this._finalReferredCallStatus = status;

// exec
await this._performReferAction({
  refer_status: this.referStatus,
  ...(this._finalReferredCallStatus && {final_referred_call_status: this._finalReferredCallStatus})
});
```
Ideally, it would be good to add a unit test for this case.

ousamabenyounes:
Thanks @davehorton — good catch. Fixed in `d060fb7`.

The final NOTIFY status is now recorded synchronously in `_handleNotify` (before the `eventHook` round trip that a racing BYE could interrupt) and included from `exec()`, so `final_referred_call_status` survives the `NOTIFY(≥200) → BYE` race regardless of which exit path wins the memoised action. The `200` threshold became a named constant (`SIP_FINAL_RESPONSE_MIN`) while touching those lines.

**Validation (RED → GREEN):** added an 8th case to `test/unit/sip-refer-action-hook.test.js` — *"the final NOTIFY status still reaches the actionHook when a BYE races the eventHook round trip"* — that holds the `eventHook` request in flight, fires the BYE, then releases it.

- RED on the previous head (`72f9ee0`): `AssertionError: final_referred_call_status must survive a BYE racing the final NOTIFY eventHook — undefined !== 200`
- GREEN on `d060fb7`: `# tests 8 # pass 8 # fail 0`

`run-ci.sh` replay (`node:20`, `jslint` + `test:unit`): baseline `origin/main` 10 pass / 0 fail → this branch 18 pass / 0 fail, no regressions, 100% line coverage on the changed lines of `lib/tasks/sip_refer.js`. The full `npm test` regression suite (docker-compose testbed + speech secrets) is CI-only as before; the CI build here is awaiting your workflow approval for the fork PR.

davehorton (approved):

PR #1581, opened 2026-08-26
fix: end the call when dial finishes with no actionHook, or a failed one

xquanluu:
- a websocket application was parked in _awaitCommandsOrHangup after a dial it was never asked about: no actionHook at all, or one whose request failed. An http application already ends the call here simply by running out of verbs; a websocket one sat in dead air until timeLimit
- TaskDial._endSessionUnlessHandedOff -> CallSession.expectNoFurtherVerbs suppresses the await for that one task iteration, so the loop falls out and the call tears down normally
- verbs still queued are unaffected: the flag is read behind the tasks.length === 0 test, so the next verb always runs
- skipped for a nested dial, and when the application already took the call elsewhere (KillReason.Replaced / ReferComplete)

### DIFF
--- a/lib/session/call-session.js
+++ b/lib/session/call-session.js
@@ -1348,6 +1348,7 @@ class CallSession extends Emitter {
     }
 
     while (this.tasks.length && !this.callGone) {
+      this._noFurtherVerbsExpected = false;  // scoped to the task about to run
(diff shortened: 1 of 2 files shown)


xquanluu:
Screenshot 2026-08-26 at 17 09 21

PR #1584, opened 2026-08-26
fix(status): don't re-send pre-transfer call statuses after a transfer

### LINKED ISSUE #1392: when a call is transferred "duplicate" status events are sent
When a call is transferred between feature servers the recieving FS sends a ringing and then an in-progess status event (because its answering the refer) but to the customers applicaiton these appear to be duplicates because the call was already answered, we should try to supress these events for a transferred call

Rehansanjay:
Fixes #1392.

## The problem

When a call is transferred between feature servers, the receiving server gets a fresh INVITE — the `context-<key>` URI handled in `lib/middleware.js`, which sets `transferredCall: true`. That session then runs its normal lifecycle: `trying` from the `InboundCallSession` constructor, then `ringing` / `early-media` / `in-progress` as it accepts the REFER.

The application was already told all of that by the server that first handled the call, so it receives a second round of status events for a call it believes was answered long ago.

## The fix

`_notifyCallStatusChange` now skips only `executeStatusCallback` for those statuses when `isTransferredCall` is true.

Deliberately **not** an early `return`: the same method also starts `record-all-calls` on `in-progress` and writes the redis call record. Returning early would silently stop recording on every transferred call. `callInfo`, the redis record and the recorder all still run exactly as before — only the application-facing notification is suppressed. `completed` is untouched, so hangup reporting is unchanged.

One judgement call worth flagging: I included `trying` and `early-media` alongside the `ringing` / `in-progress` pair named in the issue, since they are duplicates by exactly the same argument (`InboundCallSession` sends `trying` unconditionally in its constructor). Happy to narrow it to just the two named statuses if you'd prefer the smaller change.

## Local CI replay scope

What I actually ran, and what I did not.

**Ran:** `npm run test:unit` — 13/13 pass (8 pre-existing, 5 new). `npm run jslint` — clean.

**New tests:** `test/unit/transferred-call-status.test.js`. Reverting only the `call-session.js` change makes three of them fail:

```
✖ a transferred call does not re-notify statuses the first server already sent
✖ suppressing the notification still updates callInfo and the redis call record
✖ suppressing the notification still starts record-all-calls
ℹ tests 13  ℹ pass 10  ℹ fail 3
```

The other two new tests — a normal call still notifies all four statuses, and a transferred call still notifies `completed` — pass with and without the patch. They are there to catch over-suppression rather than to prove the fix.

**Not run:** the full `npm test`. I don't have drachtio / FreeSWITCH / MySQL / Redis available, so this has **not** been exercised against a real transfer between two live feature servers. That is the one thing worth confirming before merge.

Small note: the new test sets `ENCRYPTION_SECRET` itself, because `call-session.js` needs it at require time and `test:unit` passes no env.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lib/session/call-session.js
+++ b/lib/session/call-session.js
@@ -483,6 +483,24 @@ class CallSession extends Emitter {
   }
 
   /**
+   * returns true if this status was already reported to the application by the feature
+   * server that handled the call before it was transferred here.
+   *
+   * A transferred call arrives as a fresh INVITE, so this session runs through trying,
+   * ringing and answered again as it accepts the REFER. The application has already seen
+   * all of those from the original server, so sending them a second time looks like
+   * duplicate events for a call it believes is long since answered.
+   */
+  _isStatusAlreadyReportedBeforeTransfer(callStatus) {
+    return this.isTransferredCall && [
+      CallStatus.Trying,
+      CallStatus.Ringing,
+      CallStatus.EarlyMedia,
+      CallStatus.InProgress
+    ].includes(callStatus);
+  }
+
+  /**
    * returns true if this session is an inbound call session
    */
   get isInboundCallSession() {
@@ -3151,7 +3169,16 @@ Duration=${duration} `
       sipReasonHeader ?? reasonHeaderFromSipMessage(msg));
     if (typeof duration === 'number') this.callInfo.duration = duration;
     if (headers) this.callInfo.sipHeaders = headers;
-    this.executeStatusCallback(callStatus, sipStatus);
+
+    /* callInfo and the redis call record are still updated below; only the
+       application-facing notification is skipped */
+    if (this._isStatusAlreadyReportedBeforeTransfer(callStatus)) {
+      this.logger.debug({callStatus},
+        'CallSession:_notifyCallStatusChange - suppressing status already sent before transfer');
+    }
+    else {
+      this.executeStatusCallback(callStatus, sipStatus);
+    }
 
     // update calls db
     //this.logger.debug(`updating redis with ${JSON.stringify(this.callInfo)}`);
--- a/test/unit/transferred-call-status.test.js
+++ b/test/unit/transferred-call-status.test.js
@@ -1,0 +1,103 @@
+const test = require('node:test');
+const assert = require('node:assert');
+
+/* call-session decrypts credentials at require time, so it needs a secret present */
+process.env.ENCRYPTION_SECRET = process.env.ENCRYPTION_SECRET || 'foobar';
+process.env.JAMBONES_LOGLEVEL = process.env.JAMBONES_LOGLEVEL || 'error';
+
+const CallSession = require('../../lib/session/call-session');
+const {CallStatus} = require('../../lib/utils/constants');
+
+/* a CallSession is far too entangled to construct here, and none of what
+   _notifyCallStatusChange touches needs the constructor to have run */
+const makeSession = ({transferredCall = false, recordAllCalls = false} = {}) => {
+  const calls = {statusCallback: [], redis: [], callInfo: [], recorderStarted: 0, recorderStopped: 0};
+  const session = Object.create(CallSession.prototype);
+
+  Object.assign(session, {
+    callMoved: false,
+    notifiedComplete: false,
+    serviceUrl: 'http://127.0.0.1:3000',
+    application: {transferredCall, record_all_calls: false},
+    accountInfo: {account: {record_all_calls: recordAllCalls}},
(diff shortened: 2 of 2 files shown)


Rehansanjay:
Bumping this one gently — it closes #1392.

A call transferred between feature servers gets a fresh INVITE, and the receiving server re-emits trying / ringing / early-media / in-progress, which the application sees as duplicates of statuses it already handled. The fix suppresses the pre-transfer statuses on a transferred leg.

+131/-1 across two files, and nearly all of that is the regression test — the change itself is small. Green since 26 Aug. No rush, just making sure it has not fallen behind the queue.

Rehansanjay:
Re-verified against current `main` (`a666319`) — 0 commits behind, still mergeable, bug still present.

Worth flagging that this sits directly on top of #1579: the line my change replaces in `_notifyCallStatusChange` is immediately below the `sipReasonHeader ?? reasonHeaderFromSipMessage(msg)` that landed there on 21 Aug. So this is a fix inside the status-callback path as it exists now, not against an older shape of it.

The bug (issue #1392): a transferred call arrives as a fresh INVITE, so this session runs through trying / ringing / early-media / in-progress again while accepting the REFER. The application already saw all four from the original feature server, so it gets a second set for a call it believes was answered minutes ago. Only the application-facing notification is skipped — `callInfo`, the redis call record and record-all-calls all still run, which is what three of the five tests exist to pin.

**Local test replay** (jambonz runs no CI on fork PRs), at `df3dd4f` with `ENCRYPTION_SECRET=test node --test test/unit/transferred-call-status.test.js`:

```
✔ a transferred call does not re-notify statuses the first server already sent
✔ a normal call still notifies every one of those statuses
✔ a transferred call still notifies completion
✔ suppressing the notification still updates callInfo and the redis call record
✔ suppressing the notification still starts record-all-calls
ℹ tests 5   ℹ pass 5   ℹ fail 0
```

**And the same run with the fix reverted**, so the tests are shown to catch the bug rather than merely exercise the code:

```
✖ a transferred call does not re-notify statuses the first server already sent
✔ a normal call still notifies every one of those statuses
✔ a transferred call still notifies completion
✖ suppressing the notification still updates callInfo and the redis call record
✖ suppressing the notification still starts record-all-calls
ℹ tests 5   ℹ pass 2   ℹ fail 3
```

@xquanluu — you reworked `callStatusChange` in #1579, so this is your area more than anyone's. Two files, +131/−1, one of which is the test.

Rehansanjay:
Thanks for merging #1585. Rebased this one onto main on top of it — `b16f2df` — since that merge touched the same file and left this `BEHIND`.

Same shape as the one you just took: one behaviour fix in `lib/session/call-session.js`, one unit test file, no infrastructure needed to run it.

Re-ran the tests on the new base:

```
✔ a transferred call does not re-notify statuses the first server already sent
✔ a normal call still notifies every one of those statuses
✔ a transferred call still notifies completion
✔ suppressing the notification still updates callInfo and the redis call record
✔ suppressing the notification still starts record-all-calls
ℹ tests 5   ℹ pass 5   ℹ fail 0
```

And the same run with the suppression reverted, so the tests are shown to catch the bug rather than exercise the code: `ℹ pass 2   ℹ fail 3`.

The three that survive a revert are deliberate — they pin that `callInfo`, the redis call record and record-all-calls all still run when the application-facing notification is skipped. Only the notification is suppressed.

Issue #1392 if it helps: a transferred call arrives as a fresh INVITE, so this session runs through trying / ringing / early-media / in-progress again while accepting the REFER, and the application sees a second set for a call it already saw answered.

Rehansanjay:
Rebased onto `main` — #1586 merging this morning put this into `BEHIND`, which blocks a merge without showing it on the PR page.

Worth flagging because #1586 touches the same file and the same subject: it inherits `siprecRecording` onto a session that arrives by transfer. This PR is the sibling case on the notification side — a transferred call arrives as a fresh INVITE, so trying/ringing/answered are reported a second time for a call the application already saw answered. The two changes sit in different parts of `call-session.js` and rebased cleanly, no conflict.

Re-verified after the rebase: 15 unit tests pass, `eslint app.js tracer.js lib` clean, and the fix is still mutation-checked — reverting the guard so `_notifyCallStatusChange` calls `executeStatusCallback` unconditionally fails 3 of the 5 tests in `transferred-call-status.test.js`, and the two that still pass are the ones pinning that suppression does not skip the callInfo/redis update or record-all-calls.

No rush from my side — flagging only because `BEHIND` is invisible from here. cc @xquanluu

davehorton (approved):

